PR #8, opened 2016-09-13
Uc9 cleanup

JanisGailis:
The issues we talked about during the last dev meeting should be fixed now. I'd appreciate if you can skim over the diff and merge it, Norman! I'll get rid of the branch myself.


forman:
Janis, I saw your PR too late. Could you please merge master first, fix any conficts and then reissue your PR?

JanisGailis:
OK!

JanisGailis:
I believe the branch can now be pulled! The one check that fails is because there's no test for coregistration, as it is subject to a major change.

PR #99, opened 2016-12-08
Data sources, added protocol select option

kbernat:
- **CLI**, **DS** added listing available protocol for every data source in _ds info_ command
- **CLI**, **DS** added protocol selection/switching in _res open_ command
- **DS** added support for OPeNDAP protocol in OpenDataPortal (HTTP) datasource


kbernat:
This branch has conflicts that must be resolved

PR #108, opened 2017-01-03
Add the possibility to delete local data

### LINKED ISSUE #107: Ability to delete locally saved data
While working on the averaging utilities (Long Term Average, as well as temporal aggregation) it becomes apparent that I need a way to delete locally stored data to ensure that a reasonable amount of disk space is used, unless a user has explicitly chosen to save all the data being downloaded (which can potentially be a lot).
#103 #93

JanisGailis:
Odp data store can now delete local data. The feature stub has also been
added to the parent class.

resolves #107 

I implemented local data deletion as part of data source implementation, as it made sense to me. However, if this is something you don't like for some reason architecturally, I can of course do the deletion directly from where I need it!

This has been verified to work in a Notebook. I didn't write a test for it, as the data sources test is being currently skipped as 'outdated'.

### DIFF
--- a/cate/core/ds.py
+++ b/cate/core/ds.py
@@ -171,7 +171,7 @@ class DataSource(metaclass=ABCMeta):
                ``datetime.datetime`` objects.
         :return: removed number of files
         """
-        return 0, 0
+        return 0
 
     @property
     def meta_info(self) -> Union[dict, None]:
--- a/cate/core/ds.py
+++ b/cate/core/ds.py
@@ -161,6 +161,18 @@ class DataSource(metaclass=ABCMeta):
         """
         return 0, 0
 
+    def delete_local(self,
+                     time_range: Tuple[datetime, datetime]) -> int:
+        """
+        Delete locally stored data.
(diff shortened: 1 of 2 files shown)


cate/core/ds.py
```diff
@@ -161,6 +161,18 @@ def sync(self,
...
+                     time_range: Tuple[datetime, datetime]) -> int:
+        """
+        Delete locally stored data.
+        The default implementation does nothing.
+
+        :param time_range: An optional tuple comprising a start and end date, which must be
+               ``datetime.datetime`` objects.
+        :return: removed number of files
+        """
+        return 0, 0
```
mzuehlke: Why are you returning a tuple here ?

JanisGailis:
@mzuehlke Good catch! That was an artifact of a previous design. Fixed now.

JanisGailis:
@mzuehlke @forman Can you consider merging this? I need this functionality in master if I'm to eventually merge the branch that contains LTA and averaging code.

Issue #128, opened 2017-02-09
Image tile computation slow for chunked, compressed source data


forman:
It currently takes up to 0.5 seconds (!) to compute a single 220 x 220 pixels image tile for a `analysed_sst` variable from e.g. 20000714120000-ESACCI-L4_GHRSST-SSTdepth-OSTIA-GLOB_LT-v02.0-fv01.0.nc, read with xarray. It has the following encoding:

    $ analysed_sst.encoding
    {'_FillValue': -32768,
     'add_offset': 273.14999,
     'chunksizes': (1, 663, 1327),
     'complevel': 1,
     'contiguous': False,
     'dtype': dtype('int16'),
     'fletcher32': False,
     'original_shape': (1, 3600, 7200),
     'scale_factor': 0.0099999998,
     'shuffle': True,
     'source': 'D:\\EOData\\SST\\20000714120000-ESACCI-L4_GHRSST-SSTdepth-OSTIA-GLOB_LT-v02.0-fv01.0.nc',
     'zlib': True}

Here is the performance analysis output generated while computing the tile:

    PERF: >>> Tile: res_1-analysed_sst-0-jet-0.0-1.0 0 0 0
    tile "arr-res_1-analysed_sst-0/0/0/0": computed, took 4.3165 sec
    tile "tra-res_1-analysed_sst-0/0/0/0": computed, took 4.3172 sec
    tile "tra-res_1-analysed_sst-0/0/0/0": stored in cache, took 0.0000 sec
    tile "rgb-res_1-analysed_sst-0-jet-0.0-1.0/0/0/0": computed, took 4.3218 sec
    tile "rgb-res_1-analysed_sst-0-jet-0.0-1.0/0/0/0": stored in cache, took 0.0010 sec
    PERF: <<< Tile: res_1-analysed_sst-0-jet-0.0-1.0 0 0 0 took 4.322905492520917 seconds

JanisGailis:
I'm writing a test for automatic dask chunking upon opening a dataset, then I'll make a pull request for the branch, maybe this will also help in this issue, as theoretically you'd have to deal with better aligned chunking. Currently a single dask chunk that's being opened when constructing any tile woud be the whole time slice, irrespective of chunking on disk. With dask-chunked xarray dataset it would be less.

Notice also that constructing a 220x220 pixel large tile might involve ~500 redundant disk reads, in case each tile is handled separately.

[...]

PR #129, opened 2017-02-09
Open with chunking

JanisGailis:
open_xarray_dataset now checks if the uncompressed data would be too large to fit in memory with default dask chunking (lat, lon, 1) and if that's the case, smaller chunks are calculated and applied to the dataset to open.


JanisGailis:
Closes #129 

See also #128

JanisGailis:
Added a comment on why it works the way it does, added integer division where appropriate!

PR #143, opened 2017-03-06
115 add provenance info to output ds

JanisGailis:
Implemented automatic operation stamping.

Information about cate, its version, operation invoked, its version, as well as default and passed inputs will be appended to the 'history' tag of an xarray Dataset.

To stamp a dataset, 'stamp' and 'version' header attributes have to be passed upon decorating an operation as such:
@op(stamp=['return'], version='1.0')
or
@op(stamp=['named1', 'named2'], version='1.0')

Currently only xr.Dataset outputs can be stamped. But this can theoretically be expanded to other output types as well.


forman:
Hi Janis, nice work! But why not `auto_history=True`? The current `stamp=[<name>, ...]` seems redundant to me.

JanisGailis:
The interface is now:

@op(version='X.x')
@op_return(add_history=True)

@op(version='X.x')
@op_output('name1', add_history=True)

There have also been additional tests added.

forman:
Thanks!

Issue #148, opened 2017-03-09
Ops import bug


JanisGailis:
When running:

```python
import cate.ops as ops
```

We get the following stack-trace:
```shell
error: unexpected exception while loading Cate plugin with entry point 'cate_ops'
--------------------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/ccitbx/miniconda3/envs/cate_env/lib/python3.5/site-packages/setuptools-27.2.0-py3.5.egg/pkg_resources/__init__.py", line 2266, in resolve
    return functools.reduce(getattr, self.attrs, module)
AttributeError: module 'cate.ops' has no attribute 'cate_init'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/ccitbx/Development/cate-core/cate/core/plugin.py", line 85, in _load_plugins
    plugin = entry_point.load()
  File "/home/ccitbx/miniconda3/envs/cate_env/lib/python3.5/site-packages/setuptools-27.2.0-py3.5.egg/pkg_resources/__init__.py", line 2258, in load
    return self.resolve()
  File "/home/ccitbx/miniconda3/envs/cate_env/lib/python3.5/site-packages/setuptools-27.2.0-py3.5.egg/pkg_resources/__init__.py", line 2268, in resolve
    raise ImportError(str(exc))
ImportError: module 'cate.ops' has no attribute 'cate_init'
--------------------------------------------------------------------------------
```

This is probably primarily due to circular imports in some operations.

PR #149, opened 2017-03-09
Fix issue #148

JanisGailis:
Fix circular imports in operations, make sure ops:cate_init can be
invoked upon importing ops, before any attempt is made to import any of
the submodules.


JanisGailis:
Not sure why travis/pr failed. Everything seems to be fine. I'll take the chances.

Issue #46, opened 2016-09-27
Type validation that can understand Python "types" created with the typing package


JanisGailis:
isinstance hiccups on validating types from the typing package when using the cli interface. See #45

JanisGailis:
One way how to solve this would be to not type such objects in the method signature, but use something like this:
```python
@op_input('date', type='Date')
@op_input('geometry', type='Polygon')
def some_method(date, geometry, something: str = 'default'):
    pass
```

JanisGailis:
I think I have a workable solution for this.

forman:
@JanisGailis, please have a look what @mzuehlke and I wrote. We now have a better understanding, so lets discuss on Monday and combine forces! Our implementation is a little more type-safe and when defining operations, and we don't need the alias. I also like that the type classes are called "<Type>Like" and have an "accepts" method to check if any value is like <Type>. 
Cheers
    Norman

JanisGailis:
We now have complex types.

Issue #130, opened 2017-02-09
Spatial, temporal and variable selection objects


JanisGailis:
Data types with special meaning, e.g. GeoPoint, GeoBox, TimeRange, etc. for which we then create fancy input GUI components. For example, if some operation has a parameter "region" of type GeoBox, users can pick an existing region from "My Places".

See also #46 and #72 

- [x] Spatial object Point
- [x] Spatial object Polygon
- [x] Temporal selection object 
- [x] Variable selection object

JanisGailis:
Shapely seems like a good candidate for spatial objects:
http://toblerity.org/shapely/manual.html

We could use ```shapely.Point``` for lat/lon points and either ```shapely.LinearRing```, ```shapely.Polygon``` or ```shapely.MultiPolygon``` for other geometric objects.

A ```LinearRing``` is a closed collection of lines that don't cross themselves.
A ```Polygon``` is a single polygon that can have holes
A ```MultiPolygon``` is a collection of non-overlapping ```Polygon``` features

For interoperability with the GUI, the possibility to convert Shapely objects to Well Known Text [[1]](https://en.wikipedia.org/wiki/Well-known_text) [[2]](http://toblerity.org/shapely/manual.html#well-known-formats) might come in handy.

If we at some point use Shapely to handle a large amount of features, ```Shapely.speedups``` [[3]](http://toblerity.org/shapely/manual.html#performance) might come in handy.

forman:
Janis, I agree, let's go for shapely! I believe, it is a geopandas dependency anyway.

JanisGailis:
- [ ] Spatial object Point
- [ ] Spatial object Polygon
- [ ] Temporal selection object

JanisGailis:
Issue #46 has now been closed, so we have a new flexible way to define complex types for use in operations and to facilitate the integration with the GUI. This can/should now be used to create complex types to use across the board wherever spatial/time/variable selection is required.

JanisGailis:
PR #152

Issue #72, opened 2016-10-13
Allow arbitrary polygons for spatial subsetting


forman:
For spatial subsetting, users should be able to define a region of interest by providing arbitrary polygons.

Users may provide such region polygons by
- Constructing a geometry-like object (API)
- Reading from a file (API, CLI, GUI)
  - Geometry-WKT
  - Shapefile
  - GeoJSON
- Providing a WKT string (API, CLI)
- Drawing a polygon on a map or globe (GUI only)

lkeupp:
actually, this is not neccessary for UC6. 
But non-rectangular polygons will be needed in the future and the can be used for Chris's idea regarding testing of other regions for index calculation

JanisGailis:
See also #46

PR #151, opened 2017-03-10
46 type validation

JanisGailis:
Fixes issue #46 

Goes a long way towards resolving #130 and #72 

The main idea is an added core module types.py that lets us define complex types, as well as some auxiliary methods for working with these, such as validation routines, resolving routines, or anything else we might want to have. Then the types can be used as:
```python
from op.core.types import TYPE, TYPE_ALIAS, is_type, to_op_object

@op
@op_input('name1', data_type=TYPE)
def some_op(name1: TYPE_ALIAS):
   # Get the 'default' object (a List, a Shapely shape, a Pandas Dataframe)
   name1 = to_op_object(name1, TYPE) 
   # Do something
   pass
```
or more generally:
```python
def some_method(name1: TYPE_ALIAS):
    # Use instead of isinstance()
    is_type(name1, TYPE)
    # Do something useful
    pass
```

I haven't paid too much attention to code architecture, so everything is in a single file, without exported imports to core. This is because I'd like to have your opinion on where this should go and how this would fit in the overall architecture. Should it go into util, core, separate 'types' folder under cate?. Should the abstract interface class be split out and left in core, while the defintions of particular types go to the 'types' folder, as with 'ds')?

Any comments welcome!

### DIFF
--- a/cate/core/op.py
+++ b/cate/core/op.py
@@ -560,7 +560,7 @@ def op_return(data_type=None,
 
 def _update_properties(old_properties: dict, new_properties: dict):
     for name, value in new_properties.items():
-        if value is not None and (name not in old_properties or old_properties[name] is None):
+        if value is not None:
             old_properties[name] = value
 
 
--- a/cate/core/types.py
+++ b/cate/core/types.py
@@ -51,7 +51,7 @@ and add the new definition to TYPES.
 
 """
 from abc import ABCMeta, abstractmethod
-from typing import Union, Any, List, Optional
+from typing import Union, Any, List
 
 from cate.util.misc import to_list
 
@@ -155,7 +155,7 @@ _TYPES = {POLYGON: _Polygon,
 
 def is_type(value: Any, maybe_type: Any) -> bool:
     """
-    Replacement for the built in isinstance() that works with custom Cate
+    Replacement for the built-in isinstance() that works with custom Cate
     types.
 
     :param value: Value to check
@@ -186,6 +186,3 @@ def to_op_object(value: Any, typ: Any) -> Any:
     :param typ: Complex type
     """
     return _TYPES[typ]._to_op_object(value)
-
-print(is_type(0.45, VARIABLE))
-print(to_op_object('aa,bb,cc', VARIABLE))
--- a/cate/core/types.py
+++ b/cate/core/types.py
@@ -40,11 +40,20 @@ def some_op(name1: TYPE_ALIAS -> TypeNotation[TYPE]:
     # Do something useful
     pass
 
-To add a new type, add a new type definition, an alias and a validation
-function and add the new definition to _IS_TYPE.
+Or:
+def some_method(name1: TYPE_ALIAS) -> bool:
+    is_type(name1, TYPE)
+    # Do something useful
+    pass
+
+To add a new type, add a new type definition, an alias and a type class
+and add the new definition to TYPES.
 
 """
+from abc import ABCMeta, abstractmethod
 from typing import Union, Any, List, Optional
+
+from cate.util.misc import to_list
 
 # Using strings instead of ints for better debug message readability down the
 # line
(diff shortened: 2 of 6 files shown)


JanisGailis:
Seems that some tests are failing. Looking into that.

JanisGailis:
Yay!

JanisGailis:
There seem to be two related PEPs.
[PEP484](https://www.python.org/dev/peps/pep-0484/)
[PEP3119](https://www.python.org/dev/peps/pep-3119/)

```typing``` is an implementation of PEP484, it seems that it's strictly meant for static type checking, not inspection. It might be that trying to make it bend towards inspection is a serious anti-pattern.

PEP3119 on the other hand seems to be meant to be used in cases similar to our use-case.

**Comment by forman** [comment]
```
from abc import abstractclassmethod
from collections import namedtuple
from typing import Union, Tuple, Any, Generic, TypeVar
from unittest import TestCase

T = TypeVar('T')


class Like(Generic[T]):
    TYPE = None

    def __init__(self):
        raise NotImplementedError('%s cannot be instantiated' % self.__class__)

    @classmethod
    def name(cls) -> str:
        return cls.__name__

    @classmethod
    def accepts(cls, value: Any) -> bool:
        try:
            cls.convert(value)
            return True
        except ValueError:
            return False

    @abstractclassmethod
    def convert(cls, value: Any) -> T:
        pass

    @classmethod
    def format(cls, value: T) -> str:
        return str(value)


# 'Point' is an example type which may come from Cate API or other required API.
Point = namedtuple('Point', ['x', 'y'])


# 'PointLike' represents a type to be used for values that should actually be a 'Point' but
# also have a representation as a `str` such as "2.1,4.3", ar a tuple of two floats (2.1,4.3).
# The "typing type" for this is given by PointLike.TYPE.

class PointLike(Like[Point]):
    # The "typing type"

    TYPE = Union[Point, Tuple[float, float], str]

    @classmethod
    def convert(cls, value) -> Point:
        try:
            if isinstance(value, Point):
                return value
            if isinstance(value, str):
                pair = value.split(',')
                return Point(float(pair[0]), float(pair[1]))
            return Point(value[0], value[1])
        except Exception:
            raise ValueError('cannot convert value <%s> to %s' % (value, cls.name()))

    @classmethod
    def format(cls, value: Point) -> str:
        return "%s, %s" % (value.x, value.y)


# PointLike = NewType('PointLike', _PointLike)


# 'scale_point' is an example operation that makes use of the PointLike type for argument point_like

def scale_point(point_like: PointLike.TYPE, factor: float) -> Point:
    point = PointLike.convert(point_like)
    return Point(factor * point.x, factor * point.y)


class PointLikeTest(TestCase):
    def test_use(self):
        self.assertEqual(scale_point("2.4, 4.8", 0.5), Point(1.2, 2.4))
        self.assertEqual(scale_point((2.4, 4.8), 0.5), Point(1.2, 2.4))
        self.assertEqual(scale_point(Point(2.4, 4.8), 0.5), Point(1.2, 2.4))

    def test_abuse(self):
        with self.assertRaises(ValueError) as e:
            scale_point("A, 4.8", 0.5)
        self.assertEqual(str(e.exception), "cannot convert value <A, 4.8> to PointLike")

        with self.assertRaises(ValueError) as e:
            scale_point(25.1, 0.5)
        self.assertEqual(str(e.exception), "cannot convert value <25.1> to PointLike")

    def test_new(self):
        with self.assertRaises(NotImplementedError) as e:
            PointLike()
        self.assertEqual(str(e.exception), "test.core.test_types.PointLike cannot be instantiated")

    def test_name(self):
        self.assertEqual(PointLike.name(), "PointLike")

    def test_accepts(self):
        self.assertTrue(PointLike.accepts("2.4, 4.8"))
        self.assertTrue(PointLike.accepts((2.4, 4.8)))
        self.assertTrue(PointLike.accepts([2.4, 4.8]))
        self.assertTrue(PointLike.accepts(Point(2.4, 4.8)))
        self.assertFalse(PointLike.accepts("A, 4.8"))
        self.assertFalse(PointLike.accepts(25.1))

    def test_format(self):
        self.assertEqual(PointLike.format(Point(2.4, 4.8)), "2.4, 4.8")

```

**Comment by JanisGailis** [comment]
@forman @mzuehlke 
Pretty nice! 

I like the direction you're taking with TypeVar, Generic and 'Like' a lot. However, what you suggest won't solve Issue #46 as far as I can see. Consider the following:

```python
@op()
def scale_point(point_like: PointLike.TYPE, factor: float) -> Point:
    point = PointLike.convert(point_like)
    return Point(factor * point.x, factor * point.y)

class PointLikeTest(TestCase):
    def test_registered_op(self):
        registered_op = OP_REGISTRY.get_op(object_to_qualified_name(scale_point))
        registered_op("2.4, 4.8", 0.5)
```
Will fail (haven't tested, but very, very sure :) ) on ```isinstance(value, data_type)``` in cate/util/opmetainfo.py on lines 321 and 351. This is because the automatically derived type of ```point_like``` is still ```Union[Point, Tuple[float, float], str]```. That's a bit of a chicken and egg problem - if you hint the actual type object, offline type hinting will be broken, if you hint a valid hint from typing, there's no way you can get any information about the type object this 'typing type' belongs to.

What we could possibly do though is write the type classes in a way that we can hook into ```isinstance()``` [StackOverflow](http://stackoverflow.com/questions/3570796/why-use-abstract-base-classes-in-python). But then, if we want to have both - type checking at runtime, as well as type hinting in off-line mode, we still have to override what is in the ```data_type``` of a registered operation that uses the particular type. Hence:
```python
@op()
@op_input('point_like', data_type=PointLike)
def scale_point(point_like: PointLike.TYPE, factor: float) -> Point:
    point = PointLike.convert(point_like)
    return Point(factor * point.x, factor * point.y)
```

I would really like to be wrong on this one :)

forman:
> Will fail (haven't tested, but very, very sure :) ) on isinstance(value, data_type) ...

Janis, the use of `isinstance` in the validation is preliminary code, that's why we have the 

     # TODO (forman, 20160907): perform more elaborated input type checking here (issue #46)

With our own `<Type>Like` types we can perform the more elaborated input type checking by *not* using instanceof, which, btw is also not very pythonic.

JanisGailis:
Whatever you want to do for type checking will have to work on ``Union[Point, Tuple[float, float], str]```, because that's what you pass in ```def scale_point(point_like: PointLike.TYPE```, not ```<Type>Like```

JanisGailis:
Won't merge, the rationale behind this PR has been solved and implemented on master.

PR #152, opened 2017-03-14
130 new types

JanisGailis:
There are 4 new types:

PointLike
PolygonLike
Variable
DateRange

I'd appreciate your thoughts on how these look from the point of view of GUI/CLI integration. Especially the DateRange. I currently made it an explicit start/end date tuple where start has to be sooner than end. This is also how we use times across the board, so this would change all the start/end dates in signatures.

However, I remember there being talks about potentially making the time selection object a list of datetime slices instead. We can change to this later if required, but at the same time, if you feel like we should define the interface like that instead, let me know!

### DIFF
--- a/cate/core/types.py
+++ b/cate/core/types.py
@@ -93,7 +93,7 @@ class Like(Generic[T], metaclass=ABCMeta):
 
 
 # ===== Like-derived types below =====
-class Variable(Like[List[str]]):
+class VariableNamesLike(Like[List[str]]):
     """
     Type class for Variable selection objects
 
@@ -180,7 +180,7 @@ class PolygonLike(Like[Polygon]):
                     return polygon
             if isinstance(value, str):
                 value = value.lstrip()
-                if value[:7] == 'POLYGON':
+                if value[:7].lower() == 'polygon':
                     polygon = loads(value)
                 else:
                     val = [float(x) for x in value.split(',')]
@@ -189,17 +189,17 @@ class PolygonLike(Like[Polygon]):
                     return polygon
         except Exception:
             raise ValueError('cannot convert {} to a valid'
-                             ' shapely Polygon'.format(value))
+                             ' Polygon'.format(value))
 
         raise ValueError('cannot convert {} to a valid'
-                         ' shapely Polygon'.format(value))
+                         ' Polygon'.format(value))
 
     @classmethod
     def format(cls, value: Polygon) -> str:
         return value.wkt
 
 
-class DateRange(Like[Tuple[datetime, datetime]]):
+class TimeRangeLike(Like[Tuple[datetime, datetime]]):
     """
     Type class for temporal selection objects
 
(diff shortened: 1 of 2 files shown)


cate/core/types.py
```diff
@@ -86,4 +92,146 @@ def format(cls, value: T) -> str:
...
+        2. a string of datetime string start/end dates 'YYYY-MM-DD,YYYY-MM-DD'
+        3. a tuple of start/end datetime datetime objects
+        4. a tuple of start/end datetime date objects
+
+    Converts to a tuple of datetime objects
+    """
+    TYPE = Union[Tuple[str, str], Tuple[datetime, datetime], Tuple[date, date], str]
+
+    @classmethod
+    def convert(cls, value: Any) -> Tuple[str]:
```
mzuehlke: shouldn't this be `Tuple[datetime, datetime]`

cate/core/types.py
```diff
@@ -86,4 +92,146 @@ def format(cls, value: T) -> str:
...
+            if isinstance(value, Polygon):
+                if value.is_valid:
+                    return value
+            if isinstance(value, list):
+                polygon = Polygon(value)
+                if polygon.is_valid:
+                    return polygon
+            if isinstance(value, str):
+                value = value.lstrip()
+                if value[:7] == 'POLYGON':
```
mzuehlke: lower case should be allowed, too

forman:
Hi Janis,
Looks good!
A few remarks:
* when we raise `ValueError` from a `<Type>Like` class, we may avoid developer terms like `shaply`, because these messages will be seen by end users
* `Variable` may be called `VariableName` or better `VariableNamesLike` because the resulting object is *not* a variable, but a name string. If we have explicitly a single variable name, it may be `VariableName`.
* `DateRange` could be `TimeRangeLike` as it seems to be a range within time, not a range of dates. The resulting object should be tuple containing two parsed time values that could optionally have an ISO time part (Thh:mm:ss.micros). Or does *Date* refer to some special semantics? For example, if I provide the time range `["2017-01-01", "2017-01-01"]` it should be the same as ["2017-01-01T00:00:00", "2017-01-02T00:00:00"]`, which would cover any time greater than or equal to 2017-01-01 00:00:00 and less than 2017-01-02 00:00:00.

cate/core/types.py
forman: Only if WKT is case-insensitive. Is it?

forman:
Our GUI Toolkit blueprintjs only supports "date ranges": http://blueprintjs.com/docs/#components.datetime.daterangeinput. Maybe we'll stay with just dates rather than times.

cate/core/types.py
JanisGailis: After a quick search it seems that according to the standard WKT is not case sensitive. Although pretty much always all caps are used. So this is a good point.

cate/core/types.py
JanisGailis: Yes

JanisGailis:
The DateRange converts to datetime because that's what xarray uses natively:
http://xarray.pydata.org/en/stable/time-series.html

That being said, I don't think we'll actually work with anything smaller than days. Passing a '2001-01-01, 2001-01-02' results in both dates being selected fully. That is, both ends are included in the interval:

```python
    expected = (datetime(2001,1,1), datetime(2002,1,1,23,59,59))
    actual = DateRange.convert('2001-01-01, 2002-01-01')
    self.assertTrue(actual == expected)
```

The type will of course also work if one passes in datetime strings (as opposed to dates).

I didn't quite understand from the comments what's the verdict on DateRange type. Do you think it needs any changes apart from changing the name to TimeRangeLike?

PR #155, opened 2017-03-16
72 spatial subset polygon

### LINKED ISSUE #72: Allow arbitrary polygons for spatial subsetting

JanisGailis:
Spatial subset operation using the PolygonLike. Can mask Polygon exterior in case a generic Polygon is supplied. Won't spend time on trying to calculate the mask in case a simple box Polygon is provided.

@forman @mzuehlke I do some validation in the operation for the bounding box. Do you think this should move to the type instead?

Closes #72


mzuehlke:
The PR loos good to me. I cleaned up some PEP8 stuff. I will merge it to master.
The build failure seems to be related to some network issues...

Regarding the bbox validation. Your are validating the bounding box not the polygon. I would leave it here for the moment. If we see this code more often we should see.

JanisGailis:
Great, thanks!

Issue #160, opened 2017-03-22
Diff operation


JanisGailis:
For Cate demo in the cloud workshop, a diff operation (ds1 - ds2) is desired. Currently anomaly calculation is used. Create a standalone diff operation instead.

PR #161, opened 2017-03-22
160 diff op

JanisGailis:
A more robust difference operation has been implemented and tested.

Tests have been written for the ds_arithmetics operation and found bugs fixed.


JanisGailis:
Closes #160

JanisGailis:
If anyone is looking at this, wait a bit before merging, I want to add one more test for both ops.

JanisGailis:
Done

forman:
I like the diff op, but I still disagree with the arithmetic op. We should really move to an op, where users can specify a normal expression which we either 

1. `eval` without globals but with a safe locals env.
2. and/or use `numexpr` also for a performance boost.

Issue #162, opened 2017-03-23
Subset temporal with new types


JanisGailis:
Change subset temporal to use the new typing system.

PR #164, opened 2017-03-23
Subset temporal with TYPEs

JanisGailis:
Subset temporal now uses the new TYPE system. It also now works only
with datetime time coordinates and raises an error if it received
anything else.

All subset operations now also add history. Additional tests to test
new functionality, as well as functionality through the operations
registry, have been added.


JanisGailis:
Closes #162

Issue #150, opened 2017-03-09
ENSO index calculation with user defined regions


JanisGailis:
Kevin wants to be able to do ENSO Index calculation (from SST data, as in #74 ) using arbitrary, user-defined regions.

PR #171, opened 2017-03-28
74 enso index calculation

### LINKED ISSUE #74: ENSO index calculation from SST data
Provide the means to perform an [El Niño-Southern Oscillation](https://de.wikipedia.org/wiki/El_Ni%C3%B1o-Southern_Oscillation) (ENSO) index calculation from SST. This may either be provided as a standard built-in operation or as a demonstration for a user-supplied plug-in.

A more sophisticated version of this operation is the [Multivariate ENSO Index](https://en.wikipedia.org/wiki/Multivariate_ENSO_index) (MEI), see issue #75.

JanisGailis:
ENSO Indices calculation operations. Before merging this, merge #169 

Closes #74 #150


cate/ops/index.py
```diff
@@ -29,17 +29,28 @@
...
+from cate.ops.anomaly import anomaly_external
+from cate.core.types import PolygonLike, VarName
+
+
+_ALL_FILE_FILTER = dict(name='All Files', extensions=['*'])
 
 
 @op(tags=['index', 'nino34'])
-def nino34(ds: xr.Dataset, file: str, var: str, threshold: float=False):
+@op_input('file', file_open_mode='w', file_filters=[dict(name='NetCDF', extensions=['nc']), _ALL_FILE_FILTER])
```
mzuehlke: file is a very generic name

cate/ops/index.py
```diff
@@ -29,17 +29,28 @@
...
+@op_input('var', value_set_source='ds', data_type=VarName)
+def enso_nino34(ds: xr.Dataset,
+                var: VarName.TYPE,
+                file: str,
+                threshold: float=None):
```
mzuehlke: Do these operation have a return value, because here is no return type

JanisGailis: Yes, they do, thanks :)

cate/ops/index.py
mzuehlke: Id you had agreed on that, I'm fine.

JanisGailis: I don't particularly care :D

PR #175, opened 2017-04-06
153 revise data source api

kbernat:
Implemented make_local and update_local for both ESA ODP and local data sources.

1. make_local allows to copy data from one data source to other (local and remote (ESA ODP).  User can select temporal_coverage, sparial_coverage and names of variables to be copied (dataset coordinates are copied automatically)
 
2. update_local allows to extends or reduce time_coverage.


cate/ds/esa_cci_odp.py
```diff
@@ -446,81 +464,61 @@ def sync(self,
...
-            return len(outdated_file_list), len(selected_file_list)
+            self.make_local(self._master_id(), None, time_range, None, None, monitor)
         else:
-            return 0, 0
+            raise ValueError('Unsupported protocol', protocol)
+        return 0, 0
+
+    def update_local(self,
+                     local_id: str,
+                     time_range: Tuple[datetime, datetime],
```
JanisGailis: Same here, maybe it's better to do:
```python
time_range: TimeRangeLike.TYPE

time_range = TimeRangeLike.convert(time_range)
```
Then you can pass in other things that can be converted to a datetime Tuple into the method. Of course, if this method does not face user interfaces, this might be redundant.

cate/ds/esa_cci_odp.py
```diff
@@ -446,81 +464,61 @@ def sync(self,
...
 
-        for filename, _, _, _, _ in selected_file_list:
-            dataset_file = os.path.join(dataset_dir, filename)
-            try:
-                os.remove(dataset_file)
-                removed_count += 1
-            except:
-                # File busy on Windows, move on
-                pass
+    def delete_local(self, time_range: Tuple[datetime, datetime]) -> int:
```
JanisGailis: Same here

cate/ds/esa_cci_odp.py
```diff
@@ -548,14 +546,10 @@ def open_dataset(self,
                      var_names: VarNamesLike.TYPE = None,
                      protocol: str = None) -> Any:
         time_range = TimeRangeLike.convert(time_range) if time_range else None
-        # TODO (kbernat): support region constraint here
         if region:
-            raise NotImplementedError('EsaCciOdpDataSource.open_dataset() '
-                                      'does not yet support the "region" constraint')
-        # TODO (kbernat): support var_names constraint here
+            region = GeometryLike.convert(region)
```
JanisGailis: Same here as a few comments above. Not sure if ```GeometryLike``` is fully implemented. But note also, that ```GeometryLike``` should be valid even if a line is provided. Maybe ```PolygonLike``` is better for a region.

cate/ds/esa_cci_odp.py
```diff
@@ -578,22 +572,195 @@ def open_dataset(self,
...
+        :param files_description_list:
+        :param protocol:
+        :return:
+        """
+        return [file_rec[4][protocol].replace('.html', '') for file_rec in files_description_list]
+
+    def _make_local(self,
+                    local_ds: LocalDataSource,
+                    time_range: TimeRangeLike.TYPE = None,
+                    region: GeometryLike.TYPE = None,
```
JanisGailis: Same here

cate/ds/local.py
```diff
@@ -40,25 +41,29 @@
...
+from math import ceil, floor
+from typing import Optional, Sequence, Union, Any, Tuple
+from xarray.backends.netCDF4_ import NetCDF4DataStore
 
-from cate.core.ds import DATA_STORE_REGISTRY, DataStore, DataSource, open_xarray_dataset
+from cate.conf import get_config_value
+from cate.conf.defaults import NETCDF_COMPRESSION_LEVEL
+from cate.core.ds import DATA_STORE_REGISTRY, DataStore, DataSource, open_xarray_dataset, query_data_sources
 from cate.core.ds import get_data_stores_path
 from cate.core.types import GeometryLike, TimeRange, TimeRangeLike, VarNamesLike
```
JanisGailis: Same here, maybe ```PolygonLike``` is better

kbernat:
Thanks for input Janis,
Tuple -> TimeRange..  Sure! I overlooked that.. it should use new data types.
GeometryLike -> PolygonLike - OK! I see that we need some extra tests for region selection too.

/Chris

JanisGailis:
I suggest you ping Norman directly about this PR!

cate/ds/esa_cci_odp.py
```diff
@@ -443,84 +460,66 @@ def update_file_list(self, monitor: Monitor=Monitor.NONE) -> None:
...
-            return 0
+            if time_range[1] != data_source.temporal_coverage()[1]:
+                if time_range[1] < data_source.temporal_coverage()[1]:
+                    to_remove.append((time_range[1], data_source.temporal_coverage()[1]))
+                else:
+                    to_add.append((data_source.temporal_coverage()[1],
+                                   time_range[1]))
+        if to_remove:
+            for time_range_to_remove in to_remove:
+                data_source.reduce_temporal_coverage(time_range_to_remove)
```
mzuehlke: The method "reduce_temporal_coverage" seems to be missing

PR #203, opened 2017-04-24
147 1d gapfill

### LINKED ISSUE #147: Temporal 1D Gap Filling
For UC02 having the possibility to do a temporal 1D Gap Filling is desirable. It should probably just feature doing interpolation with the choice of multiple interpolation methods.

It should be implemented in a way that would work with Dask. E.g., reading the whole time-series in memory and doing something is probably not a great idea.

A good approach is probably doing xr.DataArray.rolling() over time and either manipulating the array in-place, or creating an auxiliary array that can then be used with xr.DataArrai.fillna().

JanisGailis:
A dataframe gapfilling operation that wraps pandas functionality. Is tested.

Closes #147


JanisGailis:
I'll take the liberty to merge this. It is a single operation that does not affect other parts of the code along with Flake8 fixes for master. We can always make new issues for adjusting operations as required.

mzuehlke:
Thanks for makeing the build flake8 error free 😃

PR #205, opened 2017-04-24
204 dataframe dataset

### LINKED ISSUE #204: Dataframe to Dataset conversion operation
In light of Use Case 2 it becomes apparent that there might be a need to do dataframe -> dataset conversion to employ operations that are implemented on datasets (outlier detection, plotting).

The most straightforward way to achieve this is to wrap existing xarray functionality.

JanisGailis:
Dataframe to dataset conversion utility (xarray function wrapper).

Closes #204


forman:
I prefer overloading most of our operations so that they can work on both (Geo)DataFrames and xr.Datasets. This is also more pythonic.

Issue #202, opened 2017-04-24
query() method functionality differs between remote and local data stores


JanisGailis:
A query of the remote data store returns a list of data sources that include the given pattern. A query of the local data store only returns data sources that match the pattern exactly.

See:
```python
from cate.core.ds import DATA_STORE_REGISTRY
from cate.util import ConsoleMonitor

monitor = ConsoleMonitor()

data_store = DATA_STORE_REGISTRY.get_data_store('esa_cci_odp')
local_store = DATA_STORE_REGISTRY.get_data_store('local')
soil_remote_sources = data_store.query('soil')
print(soil_remote_sources)

zemgale = '23.5, 56.3, 25.0, 57.0'
local_store.remove_data_source('local.soil.mon')
soil_remote_sources[0].make_local(local_name='soil.mon',
                                  time_range='2007-01-01, 2007-01-05',
                                  region=zemgale,
                                  monitor=monitor)
print(local_store.query('soil'))
print(local_store.query('local.soil.mon'))
```

I don't know if this is desired, a bug or an oversight and what should be the desired behavior.

PR #210, opened 2017-04-25
168 opennig local data souces

### LINKED ISSUE #168: open local data sources
CLI should use the data source ID prefix ("local.") to identify the data store it searches for the data source ID first, then search all other data stores. This means, local data stores currently cannot be used from the CLI.

Current workaround:
open any esa odp data source first and then the local one

kbernat:
closes #168, #202

### DIFF
--- a/cate/ds/local.py
+++ b/cate/ds/local.py
@@ -457,6 +457,9 @@ class LocalDataSource(DataSource):
     def name(self) -> str:
         return self._name
 
+    def matches_filter(self, name: str = None) -> bool:
+        return name.lower() in self.name.lower()
+
     def to_json_dict(self):
         """
         Return a JSON-serializable dictionary representation of this object.
--- a/test/ds/test_local.py
+++ b/test/ds/test_local.py
@@ -65,7 +65,7 @@ class LocalFilePatternDataStoreTest(unittest.TestCase):
         self.assertEqual(len(data_sources), 2)
 
         data_sources = local_data_store.query('local')
-        self.assertEqual(len(data_sources), 1)
+        self.assertEqual(len(data_sources), 2)
         self.assertIsNone(data_sources[0].temporal_coverage())
 
         data_sources = local_data_store.query('local_w_temporal')


mzuehlke:
Hi @kbernat, why implement in both relevant subclasses local and odp the same logic. It may be better to implement this in the abstract class and have no diverging implementations in the subclasses at all.

forman:
I agree, but prefer composition over inheritance.

mzuehlke:
class DataSource(metaclass=ABCMeta):
        def matches_filter(self, name=None) -> bool:
            """Test if this data source matches the given *constraints*."""
            if name and name != self.name:
                return False
            return True

    class LocalDataSource(DataSource):
        def matches_filter(self, name: str = None) -> bool:
            return name.lower() in self.name.lower()

    class EsaCciOdpDataSource(DataSource):
        def matches_filter(self, name: str = None) -> bool:
            return name.lower() in self.name.lower()

But where to delegate to ?
**And** the implementations in the subclasses can not handle `None` which is the default.

forman:
Oops, haven't seen the code. Yes, just move it up and deal with Non/empty name constraint.

kbernat:
I thought that exact name comparison is still "default behavior", but both ODP and local should act more loosely.  Now I see that there no such requirement.
@mzuehlke why DataSource.matches_filter() why we need None, and why then it returns True?

forman:
The idea was that `name` is a constraint. If it is not given, we have no constraint, thus we want an unconstrained result which is all data sources.

kbernat:
If there is no name provided then we don't run filter anyway.
```
    def query(self, name: str = None, monitor: Monitor = Monitor.NONE) -> Sequence['DataSource']:
        self._init_data_sources()
        if name:
            result = [data_source for data_source in self._data_sources if data_source.matches_filter(name)]
        else:
            result = self._data_sources
        return result
```
Anyway, I will move implementation to DataSource but will left None handling as is now in DataSource class (True when name is None).

Issue #103, opened 2016-12-14
Temporal aggregation (averaging)


JanisGailis:
We need a possibility to perform temporal averaging of data in order to go from daily to monthly datasets.

#71

PR #215, opened 2017-04-25
213 new lta

### LINKED ISSUE #213: Long Term Average overhaul
Currently Long Term Average operation works more like an independent tool, and it doesn't work with the new data sources implementation at all.

It has to be changed so that it work on datasets instead of data sources. GroupBy-Combine is a good candidate.

JanisGailis:
Overhauled long term averaging operation.

The new operation works on datasets and produces CF compliant climatological datasets.

Closes #213 #103


JanisGailis:
It seems that it's just web tests failing!
I'd appreciate if someone can merge this tomorrow!

JanisGailis:
Making the merge!

PR #258, opened 2017-06-20
236 uc6 notebook

JanisGailis:
Use Case 6 happy path notebook and scripts.


JanisGailis:
I'll take the liberty to merge this, as the changes only affect scripts and notebooks.

PR #266, opened 2017-06-26
Add matplotlib plotting capabilities. 

gunbra32:


forman:
Most code has been taken over by copy & paste. Input validation applied. Code changes of global `matplotlib` configuration has been dropped as it affects Cate'S API, CLI, and GUI which is not wanted. 

We now have two more plotting operations. Thanks, Gunnar!

Issue #238, opened 2017-05-22
Many operations are lacking sanity checks


forman:
### Expected behavior

A lot of our Python operations are lacking sanity checks and produce confusing stacktraces.

For example, all of our Python operations should raise `ValueError` with a dedicated message if a mandatory value is missing i.e. `None` has been passed to a non-nullable parameter. 

### Actual behavior

Unrelated exceptions are raised because within most operation's implementation code, illegal values, often `None` values are accessed without checking, or they are passed into lower level API where tey produce exceptions out of our control thereby increasing the length of the stacktrace.

This produces very confusing error messages for API/CLI/GUI users.

Implementation hints: 
1. Utilize `Like.convert(nullable=True)`, 
2. in package `cate.core.op` provide function `validate_op_input(op_name, input_name, actual_value)` which has access to an input's property passed to `@op_input`

The selected approach should be documented in the developer guide (#77).

### Steps to reproduce the problem

Call e.g. `pearson_correlation` with any of the mandatory inputs set to `None`.

### Specifications 

Cate 0.8.0rc2

PR #275, opened 2017-06-30
partial fix for #238

forman:


JanisGailis:
Splendid! Thanks, I've checked out how it works, can be merged now.

PR #289, opened 2017-07-10
201 jg improve correlation

### LINKED ISSUE #201: Improvements to correlation
- [x] Explore if it's possible to do without lon/lat looping in Python
- [x] Change return type for a single correlation.
- [x] Single timeseries broadcasting, see #262 
- [x] Make the operation polymorphic, see #232 
- [x] Sanity checks, see #238

### LINKED ISSUE #262: Pearson correlation single timeseries broadcasting
Currently Pearson correlation supports correlation calculation between datasets of the same shape. E.g., if two lat/lon/time datasets are provided, they must be of the same shape and then correlation will be calculated for each spatial point against the corresponding spatial point in the other dataset.

If 1D datasets of 'time' dimension are provided, then the correlation calculation will produce a single correlation coefficient for the given pair.

However, in Use Case 6 one should be able to produce a correlation map where each spatial point in a given region is correlated against the same ENSO Index timeseries. E.g., the ENSO index timeseries should be broadcast against a lat/lon/time dataset, and produce a map of correlation coefficients.

### LINKED ISSUE #273: pearson_correlation too slow
### Expected behavior

The `pearson_correlation` operation should execute fast as it uses simple, linear maths.

### Actual behavior

It is way too slow. Computing of average size lat/lon images of a few megapixels can easily take minutes.

Run use case 9.

### Specifications 

0.8.0rc7.dev2

JanisGailis:
Improvements to correlation, including a major performance improvement, updated operation development guide, updated UC9 scripts and notebook.

Closes #201 
Closes #262 
Closes #273 

@forman, @mzuehlke can you please check at least two things:
1. Update to CHANGES.md
2. Output notation of ```pearson_correlation_simple```. There is none now, I tried ```dict```, but it would complain
```shell
ValueError: output 'p_value' for operation 'cate.ops.correlation.pearson_correlation_simple' must be of type 'dict', but got type 'float64'
```


JanisGailis:
Good catch, fixing it now!

forman:
1. `pearson_correlation_simple`:  Note we have no support for multi-output ops in the GUI. Therefore we cannot use with Cate Desktop. Regarding its name, I'd prefer `pearson_correlation_scalar`.
2. `pearson_correlation_map`: Can we please just leave the `pearson_correlation` as it was before. *Map* is something cartographic, I can't see why a map has something to do with correlations.

JanisGailis:
Sure. Do you have any comments about the return type notation in ```pearson_correlation_simple``` (or scalar)?

JanisGailis:
@forman Regarding not being able to use ```pearson_correlation_scalar``` on the GUI.

Do you plan to implement something before v1.0? I believe that the 'final' (haha, makes me grin myself) form of the ```_scalar``` operation would be like it is - with multiple named ```float``` return values. It makes the most sense.

However, if you don't see yourself getting round to this as there are many *much* more important things to still achieve, I can of course change the return type to ```pd.DataFrame``` in which case it can be used on the GUI.

Let me know which one do you prefer!

The second option would also solve the notation question I made above.

forman:
I prefer the `pd.DataFrame` for time being

JanisGailis:
Great, I'll change it accordingly then and merge if there are no other notes!

PR #290, opened 2017-07-11
Get progress from xarray / dask tasks

### LINKED ISSUE #282: Using Cate monitor with deferred processing
### Expected behavior

Any operation that has the potential of running for some time should accept a monitor to give feedback about progress and provide the ability to cancel it.

### Actual behavior

When using xarray ufuncs and arithmetic capabilities, the actual processing is deferred until data is actually accessed - plotted, printed, explicitly accessed through ```.values```. Currently there is no way how to inject Cate monitor into xarray/dask land to perform monitoring, or cancel the task.

### Specifications 
Cate 0.9.0.dev3

### LINKED ISSUE #288: Implement monitor in coregistration
### Expected behavior

Co-registering two datasets should present progress information and be cancellable

### Actual behavior

Co-registration operation doesn't use a monitor

1. Run co-registration

### Specifications 

0.9.0.dev4

mzuehlke:
This adds a `dask.Callback` handler that translates the callbacks emitted from the `dask.scheduler` into progress for the `cate.Monitor` and allows to cancel the tasks.

I've used the new `DaskMonitor` in the `subset` and  `coregister` op to show the usage and test its usability.

There are some more ops that would profit from using the `DaskMonitor`. Whenever a `xarray` function invocation results in a `dask` get or compute function call. 

fixes #282
Closes #288


JanisGailis:
This is very exciting! I'll take a look and test how it works!

forman:
What about:

    monitor = parent_monitor.child(1)
    with monitor.observing("resample time slice"):
        result = resampling.resample_2d(np.ma.masked_invalid(arr_slice.values),
        ...

forman:
Regarding the 

    if monitor.is_cancelled():
        raise InterruptedError

... would be nice if Monitor impls could raise this on their own when a method is called but they are already cancelled. Very useful in `Monitor.progress()`.

forman:
As just discussed: `Cancellation(Exception)` to distinguish between interruption errors like timeouts, kill -9, etc and true cancellation by users.

Issue #280, opened 2017-07-05
Limit possible operation tags


forman:
### Expected behavior

Operations should only use tags from a limited and agreed-on set.

- [x] aggregate
- [x] anomaly
- [x] arithmetics
- [x] coregistration
- [x] correlation
- [x] index
- [x] io
- [x] outliers
- [x] pandas
- [x] plot
- [x] resampling
- [x] select
- [x] subset
- [x] timeseries
- [x] utility

### Actual behavior

We have 20+ op tags. See dropdown list in GUI.

### Specifications 

0.9.0.dev3

JanisGailis:
Can you provide the initial input to such a set? As in, which tags it makes sense to have from user's perspective when using the GUI?

forman:
When I was young, I wrote https://github.com/CCI-Tools/cate-core/wiki/Operation-Tags

JanisGailis:
That's a good initial set. I'll use those tags and raise the question if there is an operation that doesn't fit well within that set. So, according to that list 'pearson_correlation_*' would only have the ```utility``` tag.

forman:
Plus the `correlation` tag, because it is its module name.

forman:
Ah, but we could add that automatically - right?

JanisGailis:
It could be done automatically, and if it's done it should be done automatically. I'm just wondering if we actually want to put module names as tags. I mean, it makes sense to have them. But then, there are currently 16 operation modules in cate/ops. Add 7 tags mentioned in the wiki and we're back to 20+.

forman:
You're right. So let's *not* auto-tag.

JanisGailis:
I'm going over all operations to check if the tags used are not too extensive. E.g. each operation might have a module tag (not necessarily), and maybe a tag from the list in the wiki.

forman:
@JanisGailis any news here?

JanisGailis:
@forman Moving it to paused must have been done by mistake. As far as I'm concerned, this is done. I've gone over all operations and limited the tags to what we have in wiki + (maybe) module name.

forman:
Looks ok now:

![image](https://user-images.githubusercontent.com/206773/30736251-4c52b23a-9f82-11e7-8542-d130bae1aea3.png)

PR #293, opened 2017-07-11
292 jg autotags

### LINKED ISSUE #292: Automatically add module name to 'tags' list of an operation
### Expected behavior

Operations should have a tag that is the same as it's module name. E.g. 'correlation' for all operations in the correlation.py module.
See also #280 

### Actual behavior

When ```@op()``` is invoked, module name gets added to the tags list automatically. Only additional tags should be added by hand as required.

### Specifications 
Cate 0.9.0dev3

JanisGailis:
Introduce automatic operation tagging with the module name.

Closes #292

### DIFF
--- a/cate/core/op.py
+++ b/cate/core/op.py
@@ -397,8 +397,10 @@ def op(registry=OP_REGISTRY, **properties):
         op_registration = registry.add_op(op_func, fail_if_exists=False)
         try:
             default_tags = op_registration.op_meta_info.header['tags']
-            properties['tags'] = default_tags + properties['tags']
+            if default_tags[0] not in properties['tags']:
+                properties['tags'] = default_tags + properties['tags']
         except KeyError:
+            # The user has not provided any tags
             pass
         op_registration.op_meta_info.header.update({k: v for k, v in properties.items() if v is not UNDEFINED})
         return op_registration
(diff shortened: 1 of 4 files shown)


forman:
Good! However, I don't like the expression `if default_tags[0] not in properties['tags']` in the `op` decorator. First, what are `default_tags`? Second, we assume there is only one and it is the module's name, so we actually mean `if module_name not in properties['tags']`. 

I must confess, I'm currently struggling whether auto-tagging is really a good approach at all. I tend to say no. Marco too.

But implementation is great!  :-)

JanisGailis:
```default_tags``` would be tags created during function introspection. This is of course only the module name. However, it must be a list of strings to be iterable by string, not by character later on wherever it is used. Hence the notation. It could be of course spelled out. ```module_name = default_tags[0]```.

Haha, yes, I'm not sure about that myself, see my last comment  on the other issue (https://github.com/CCI-Tools/cate-core/issues/280). But then we have to make a decision on the set of tags that we use and/or the strategy of using/not using module name as part of tags.

forman:
Are you ok if we drop aut-tagging?
If so, I owe you some beers, because I brought up the idea.

JanisGailis:
Sure, it took like 20 minutes anyway! But then please comment in #280 in the strategy on using/not using module names in tag list!

PR #333, opened 2017-08-22
269 kb empty local data sources

kbernat:


mzuehlke:
Hi @kbernat , 

as now `make_local` may return `None` the return type should now be `Optional[DataSource]`.
Also, the places where `make_local` is called may have to check for `Ǹone`

kbernat:
@mzuehlke, thanks for review. I will apply all guidelines!

PR #389, opened 2017-09-22
36 jg multidimensional coregistration

### LINKED ISSUE #36: Coregistration should be able to handle datasets with more dimensions than lat/lon/time
Currently coregistration works well only if both datasets to coregister have lat/lon/time dimensions. This should not be the case, as it should be possible to bring two datasets on the same spatial lat/lon grid irrespective of how many other dimensions there are.

For example, having two datasets with such dimensions:

lat/lon/time

lat/lon/time/layer/pressure,

it should be possible to resample the second dataset to the lat/lon resolution of the first one such that it still preserves the other dimensions. E.g., instead of only 'group_by'  time, but somehow other dimensions as well.

### LINKED ISSUE #348: Upsampling failed
### Expected behavior

A slave dataset with spatial resolution lower than the master dataset should be resampled to the spatial resolution of the master

### Actual behavior

Error message 'ValueError: O3_du data array of slave dataset is not valid for coregistration. Expected coordinates are (lat, lon, time), received coordinates are ('time', 'layers', 'lat', 'lon'), consider running select_var and/or normalize operations first.'

I tried to select a variable and normalize the ds - still get same error message. Cannot figure out how to select an individual layer within a 3D dataset.

1. Open ozone dataset as resource
2. Open cloud dataset as resource
3. Apply operation 'coregister' with ozone as slave and cloud as master; use defaults.

Note: downsampling (using ozone as master and cloud as slave) works fine. Further, if the newly resampled cloud ds is used as a slave, and the old cloud ds is used as the master, the operation raises no errors (but does not show data) - implication is that this fails with datasets with a z-dim.

### Specifications 

Cate 0.9.0-dev6

JanisGailis:
Updated coregistration implementation. The interface stays the same, but coregistration can now handle:
* N-dimensional variables by doing recursive group-by operation to retrieve and then resample lat/lon slices
* Two dimensional lat/lon datasets can now be used both as master and as slave datasets

The new functionality has been tested, old tests pass, and UC09 script which uses coregistration works as expected.

Closes #36 
Closes #348


JanisGailis:
Woops

Issue #318, opened 2017-08-04
Regional subset crossing anti-meridian


HelenClifton:
### Expected behavior

Handling of international date line needs to be discussed with scientists (champion users). Is it a requirement to be able to subset over international date line ?
Once user requirements are known, then need to discuss how this is handled in cate.

### Actual behavior

Currently when region over date line is selected, cate will switch over the longitudes:
start longitude ---> end longitude
end longitude ---> start longitude 
and a region in the opposite side of the globe is selected.

JanisGailis:
@HelenClifton Can you please provide the coordinates used for selection? Cate should handle subsets over international date line.

HelenClifton:
test

HelenClifton:
@JanisGailis Here are two procedures to try

1.	Open GUI

2.	In the remote data sources list, select esacci.CLOUD.mon.L3C.CLD_PRODUCTS.multi-sensor.multi-platform.ATSR2-AATSR.2-0.r1 and click “download and/or open remote dataset” button
In “Download Data Source” window:
•	Select Time Constraint: 2007-01-01 to 2007-04-01
•	Select Region Constraint: Longitude from 130 to -130, Latitude from -30 to 30
•	Deselect “Download and make local data source“
•	Select “Open dataset from remote (in memory, via OPeNDAP)“ 
•	Click “Open Remote“


![hc 055 1](https://user-images.githubusercontent.com/29569434/29116726-b39d68ec-7cf3-11e7-8e2f-612c5b5ceaaa.PNG)


3.	Inspect World view

and

1.	Open GUI

2.	In the remote data sources list, select esacci.CLOUD.mon.L3C.CLD_PRODUCTS.multi-sensor.multi-platform.ATSR2-AATSR.2-0.r1 and click “download and/or open remote dataset” button
•	In “Download Data Source” window:
•	Select Time Constraint: 2007-01-01 to 2007-03-01
•	Leave Region Constraint and Variable Constraint unselected
•	Select “Download and make local data source”
•	Leave unique ID as default value
•	Click “Download“

![hc 055 2](https://user-images.githubusercontent.com/29569434/29116540-02220190-7cf3-11e7-9b47-7c9f4b2fe80a.PNG)


2.	Wait for download to complete then click “Open Local Dataset”

3.	In “Open Local Dataset” window:
•	Leave Time Constraint unselected
•	Select Region Constraint 20 to -20, Latitude from -30 to 30
•	Leave Variable Constraint unselected
[...]

PR #396, opened 2017-09-22
340 jg lta fails

### LINKED ISSUE #340: Either temporal_aggregation or LTA fails
### Expected behavior

[Happy Path](https://github.com/CCI-Tools/cate-core/blob/master/scripts/uc06.sh) for UC06 should still work when one selects another period, but still uses the same dataset as in the happy path.

### Actual behavior

At some point from the temporal_aggregation to the LTA cate runs into an error.

Please stick to the shell-script outlined [here](https://github.com/CCI-Tools/cate-e2e/blob/master/scripts/UC6/UC06_S2_v001.sh).

The LTA step returns:
"Long term average operation expects a monthly dataset running temporal aggregation on this dataset beforehand may help."
But within the script a temporal_aggregation step is calculated before.

### Specifications 

CLI, Linux, cate-0.9.0_dev5

JanisGailis:
* Use global temporal attributes to check temporal resolution in aggregation operations
* Adjust temporal attributes automatically upon reading a dataset
* Refactor 'normalize' operation implementation logic to util/opimpl
* Adjust temporal attributes automatically upon DatasetLike.convert
* Refactor 'subset' operation implementation logic to util/opimpl

Closes #340 
Goes a long way in solving #318

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,9 @@
 ## Changes in version 1.0.0.dev1 (unreleased)
 
 ### Improvements and new Features
+* Global temporal attributes are adjusted automatically when opening new datasets
+* Global temporal attributes are adjusted automatically when converting from data frames
+* Normalization and subsetting operation implementation logic is refactored out to util so that it can be re-used throughout Cate
 
 ### Issues Fixed/Resolved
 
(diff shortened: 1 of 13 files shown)


JanisGailis:
Appveyor test fails although there's no difference between the feature branch and master in unit test code there. I'll go ahead and merge anyway.

mzuehlke:
Hi @JanisGailis 

The failing tests in Appveyor are related to this PR. 😢 
in `adjust_temporal_attrs_impl` you create a copy of the ds: `ds = ds.copy()` before you alter some attributes. But then the close operation on the ds in the test does not work anymore because the copy does not propagate the `file_obj`, so the file is not closed and on Windows, the file can not be removed.

I have no good idea how to fix this....

JanisGailis:
Cool, so the Unit test caught something :)

I'll also think about how to solve this.

JanisGailis:
@mzuehlke I managed to fix the Appveyor build by implementing a context managed file handler that ignores the OSError when on Windows. I had used that approach in operation tests when I'm writing something to disk. The approach itself is largely borrowed from xarray people.

PR #423, opened 2017-09-25
391 nf unique resource names

forman:


forman:
@JanisGailis pls review my changes in `cate.core.op` - added new @op property `res_prefix`. Used in `cate.ops.io` and `cate.ops.plot`.

Oops, just found a "problem": the kwarg is `res_prefix` but the property assign is `resource_prefix`

forman:
@kbernat Please note that I've changed the `DataStore.query(id=None, ...)` into `DataStore.query(ds_id=None, ...)` and consistently used that first argument **with its keyword**.

Furthermore I've moved `cate.core.ds.gete_data_stores_path()` to `cate.conf.conf`

PR #430, opened 2017-09-26
Add more monitors in ops

JanisGailis:
What the title says.

### DIFF
--- a/cate/ops/correlation.py
+++ b/cate/ops/correlation.py
@@ -70,6 +70,7 @@ def pearson_correlation_scalar(ds_x: DatasetLike.TYPE,
     :param ds_y: The 'y' dataset
     :param var_x: Dataset variable to use for correlation analysis in the 'variable' dataset
     :param var_y: Dataset variable to use for correlation analysis in the 'dependent' dataset
+    :param monitor: a progress monitor.
     :returns: {'corr_coef': correlation coefficient, 'p_value': probability value}
     """
     ds_x = DatasetLike.convert(ds_x)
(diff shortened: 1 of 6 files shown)


JanisGailis:
@mzuehlke You're the expert on observing monitors! Can you please check this out? I felt a bit unsure in places if what I'm doing is the correct thing to do. Especially in correlation (monitor.observing encompasses a lot of code) and in index.py and timeseries.py.

mzuehlke:
I only looked at the operators you had modified. And improved the code in some places.
`_pearsonr` I looked a bit more closely by making sure the input is chunked and the then look at the output of the `RecordingMonitor` and by locally enabling `_DEBUG_DASK_PROGRESS` in "monitor.py".

cate/ops/correlation.py
```diff
@@ -251,8 +253,14 @@ def _pearsonr(x: xr.DataArray, y: xr.DataArray, monitor: Monitor) -> xr.Dataset:
         # deferred processing.
         # Comparing with NaN produces warnings that can be safely ignored
         default_warning_settings = np.seterr(invalid='ignore')
-        r.values[r.values < -1.0] = -1.0
-        r.values[r.values > 1.0] = 1.0
+        with monitor.child(1).observing("task 1"):
+            negativ_r = r.values < -1.0
```
JanisGailis: @mzuehlke 
So, we only monitor when we try to access the values? As in, processing from lines 239-247 would be deferred until then?

mzuehlke:
I will in a calm moment try to make the dask progress monitor a bit smarter. Because I find this multiple observing calls not the nicest solution.....

JanisGailis:
So, for future reference - as it is now, we can only wrap a single line in ``with monitor.observing():`` right?

mzuehlke:
Currently: yes

PR #442, opened 2017-09-29
removed DataStore/Source methods not reached by production code

forman:
Please make sure the code is not needed somewhere else. I tend to remove all unused code before the 1.0. I'd also like to continue consolidating the DataStore/Source APIs so there may be more to come, see TODO in `cate.core.ds` code.


kbernat:
I'm gonna to merge it. Otherwise new conflicts will appear.

PR #449, opened 2017-10-03
444 nf revise pyramid layout

forman:


forman:
See cate-desktop PR with same name

Issue #58, opened 2016-09-30
CLI (and WebAPI) take too long to launch


forman:
Both the CLI (and the WebAPI service) import _all_ package dependencies the CCI Toolbox has. This is because all operations in `ect.ops` are also imported which import packages such as `cartopy`, `basemap`, `matplotlib`, `pandas`, `xarray` which again import their own dependencies.

Both the CLI (and possibly the WebAPI service) must be modularized so that individual commands/actions have their own modules which will import only the desired packages. This is more important for the CLI, because the WebAPI is usually launched only once and remains in memory as a demon process. However, unit-level tests that require starting/stopping the service currently execute very slow due to the reasons stated above.

mzuehlke:
It seems at least in some part to be a problem of python itself. The startup time has more than doubled between python 2 and python 3. 

An interesting article (no solution included 😞 ):
[Reducing Python's startup time](https://lwn.net/Articles/730915/)

forman:
Let's refactor out the web-service instantiation from CLI unit-tests and use a mock instead.

forman:
Started this because of #465

forman:
Closing this, as there is currently no much more we can (easily) do.

PR #467, opened 2017-10-30
58 nf speedup cli

forman:
This partly addresses #58, as `cate --version` and `cate -h` now perform much faster.

Performance improvement comes from two optimisations:

1. `cate.util` does not aggregate sub-packages anymore. We have to explicitly import now. This is because some sub-packages imports are slow, notably the new `cate.util.opimpl`.   
2. CLI commands in `cate.cli.main` now only import required modules when they are executed.


forman:
@JanisGailis we prefer relative imports over absolute ones because they do not force loading the parent package's imports from `./__init__.py`.

PR #494, opened 2018-01-08
added new compute(ds, expression) operation (resolves #167)

### LINKED ISSUE #167: Arithmetics operation update
@forman: I like the diff op, but I still disagree with the arithmetic op. We should really move to an op, where users can specify a normal expression which we either

* eval without globals but with a safe locals environment, see `cate.util.safe_eval()`
* and/or  use numexpr for boosting performance.

forman:


JanisGailis:
Looks good!

PR #501, opened 2018-01-10
changed all own error classes to use the Python 3 extended raise stat…

forman:
…ement: "raise my_error() from other_error"

### DIFF
--- a/cate/core/ds.py
+++ b/cate/core/ds.py
@@ -398,40 +398,27 @@ class DataStoreRegistry:
 DATA_STORE_REGISTRY = DataStoreRegistry()
 
 
-# noinspection PyArgumentList
 class DataAccessError(Exception):
     """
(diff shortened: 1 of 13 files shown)


forman:
@kbernat, @mzuehlke, @JanisGailis 

Please review. This PR is about the Python 3 `raise` statement with `from` which we haven't used at all so far.

@kbernat please also have a look at `DataAccessError` class and its usages. which is much simpler now.

Last one to successfully review should merge please, thanks!

JanisGailis:
Should we use a similar approach in ops?

forman:
@JanisGailis you mean, introduce an `OperationError`? Only if we (must) handle this error type in specific way.

JanisGailis:
OK, I didn't realize what the main reason for having own error classes was! No, I don't think we need an ``OperationError`` at the moment then.

PR #513, opened 2018-01-26
optimized polygon conversion in _crosses_antimeridian()

forman:


bcdev:
@JanisGailis could you please review.

forman:
For this type of test, using exterior only should be necessary and sufficient. Or is there an an example of a polygon that would fail the test but still crosses the antimeridian with its interiors only?

JanisGailis:
You're right, this is just a test! We still use the original polygon afterwards.

Issue #86, opened 2016-10-21
save animated figures


lkeupp:
#85

JanisGailis:
Same question as for #85. Is it essential to have the animation for the API and CLI interfaces, or is it enough if we can do this in GUI? @forman

Issue #473, opened 2017-12-04
UC19 Happy Path


JanisGailis:
Create the demonstration scripts for Linux and Windows based on the 19th Use Case.

PR #521, opened 2018-02-12
86 jg animation

JanisGailis:
Don't merge this!

@forman, can you please provide feedback on this?

I've taken a stab at creating animations. There were multiple options around, but I chose using pure matplotlib functionality to not increase the dependency footprint.

The animation operation returns a stand alone HTML, as well as can save the animation into it. Matplotlib also allows for saving video formats and gifs, but that functionality requries ffmpeg, avconv and imagemagick installed on the host system. The HTML thing doesn't have external dependencies and it creates an animation with controls that can be used to slow it down, stop it, loop it, etc.

Works in a notebook as:
```python
import cate.ops as ops
from IPython.core.display import display, HTML
cc = ops.open_dataset('local.CLOUD_2008')
display(HTML(ops.animate_map(cc, var='cfc_low')))
```

As it returns HTML, I hope it should be easy enough to create a viewer in cate-desktop for it.

Haven't benchmarked it, but the operation shouldn't be memory hungry, as it only accesses slices one by one to create the animation. So it should work fine with large out of memory arrays too. I haven't yet added a monitor, but that should be straightforward enough.

The feedback I'd like to get is:
1. Is the general approach of using Matplotlib native stuff and HTML output fine?
2. Should the animation ops go into plot.py, or should we have them separate?
3. There are parts of plot operations that will overlap with parts of animation operations, do you think it should be refactored to have a bunch of helpers, or do you think code duplicates are fine here?
4. Any other thoughts on this?

Pending your approval, I plan to make an animation for the other plot method as well to allow generic animations to be made, implement monitors, try to make the operations cancellable and write some tests.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,6 +17,8 @@
   `read_geo_data_frame`.
 * CLI now launches a lot faster, e.g. try now `cate -h`
   [#58](https://github.com/CCI-Tools/cate/issues/58)
+* Cate can now produce animated figures
+  [#86](https://github.com/CCI-Tools/cate/issues/86)
 
 ### Fixes
 
@@ -27,8 +29,10 @@
   [#466](https://github.com/CCI-Tools/cate/issues/466)
 * Region constraint'-option for AEROSOL dataset returns 'code 20' error
   [#462](https://github.com/CCI-Tools/cate/issues/462)
- * Address problems of a user working with Cloud and Aerosol
+* Address problems of a user working with Cloud and Aerosol
   [#478](https://github.com/CCI-Tools/cate/issues/478)
+* Most projections not working in plot operations
+  [#524](https://github.com/CCI-Tools/cate/issues/524)
 
 ## Version 1.0 (10.10.2017)
 
--- a/cate/core/types.py
+++ b/cate/core/types.py
@@ -165,6 +165,26 @@ class Arbitrary(Like[Any]):
         return str(value)
 
 
+class HTML(Like[str]):
+    """
+    Represents HTML string
+    """
+    TYPE = str
+
+    @classmethod
+    def convert(cls, value: str) -> str:
+        """
+        Return **value**
+        """
+        return value
+
+    @classmethod
+    def format(cls, value: str) -> str:
+        if value is None:
+            return ''
+        return value
+
+
 class Literal(Like[Any]):
     """
     Represents an arbitrary Python literal.
--- a/cate/ops/__init__.py
+++ b/cate/ops/__init__.py
@@ -54,7 +54,7 @@ from .correlation import pearson_correlation_scalar, pearson_correlation
 from .normalize import normalize, adjust_temporal_attrs, adjust_spatial_attrs
 from .io import open_dataset, save_dataset, read_object, write_object, read_text, write_text, read_json, write_json, \
     read_csv, read_geo_data_frame, read_netcdf, write_netcdf3, write_netcdf4
-from .plot import plot_map, plot, plot_dataframe
+from .plot import plot_map, plot, plot_data_frame
 from .animate import animate_map
 from .resampling import resample_2d, downsample_2d, upsample_2d
 from .subset import subset_spatial, subset_temporal, subset_temporal_index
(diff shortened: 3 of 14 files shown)


forman:
> Is the general approach of using Matplotlib native stuff and HTML output fine?

In general, it is a good idea to not introduce new dependencies. But to answer your question, it really depends on the animation operation's output format and quality. In the SoW and URD you'll find that users want MPEG4 export of animations.

For Cate Desktop, it would be good to know when we have a plain HTML output so we can render it directly. If the output is just of type `str`, we actually don't know what to do with it unless we sniff into the content. I'd like to avoid that.

We already have a special treatment for resource objects of type `Figure`. In this case, data will be rendered in the GUI as plot.

> Should the animation ops go into plot.py, or should we have them separate?

Fine, if you call it `plot_animated_map` or so and if they could return a `Figure`. Otherwise I'd place them in `animation.py` .

> There are parts of plot operations that will overlap with parts of animation operations, do you think it should be refactored to have a bunch of helpers, or do you think code duplicates are fine here?

Always try refactoring out code duplicates to allow for easier testing and increasing code coverage.

> Any other thoughts on this?

It would be cool if animation operations could be configured to produce an MPEG stream that could be rendered in the GUI or sent to an output file. It would be even cooler if we would render the animation stream directly onto the 3D globes.

JanisGailis:
I haven't found a way how to make it work with a Figure. If I return the figure, it always shows the last frame of the animation both in cate-gui and Jupyter. Or the first frame, if the animation generation hasn't been done beforehand. In pure Python if one just creates a figure and an animation, then invoking plt.show() will present a figure in which, based on the delay, each frame is generated on the fly. So, I'd say that's not really a good way how to go about this. I much prefer an animation operation that can be observed and when finished presents a 'delay-free'  animation.

It's a string currently because I don't think we have a 'generic HTML' type I could use.

Using MPEG-4 is very much possible. But that means that ffmpeg (available on Windows, Linux and Mac OS) has to be bundled with Cate. I'm not sure we want to do this and how easy/not easy it is.

I'll make them part of separate animations.py then and proceed with monitors and stuff. I'll see if a convenient approach of refactoring presents itself. I'll currently make them work as is with HTML outputs, but it's not a problem to convert that to MPEG4 output later on, if we decide on bundling ffmpeg with Cate.

Yes, we can have a chat later this week. Please suggest a time for a devtelco, currently most slots are open for me.

JanisGailis:
If we include ffmpeg in the installation, we can also then generate an HTML5 video tag. Which can then be used to embed the video in cate-desktop in the same way as it would work now. maybe even it is then possible to use the HTML5 video tag view to export the video from the GUI to a video file.

forman:
We may introduce a new type VideoLike that could be the return type of `make_animation`. Resources of this type can be converted to a special WebAP URL which can then be rendered using the HTML5 video element. The operation can also have an optional `out` (file) argument so it can write an MPEG file.

https://developer.mozilla.org/de/docs/Web/HTML/Using_HTML5_audio_and_video

JanisGailis:
We can do that, but in that case each operation returning that type should actually create video files, but that is possible only if ffmpeg or avconv is installed on the host system.

The output of the to_jshtml() method is more like a GIF, where each frame is encoded as PNG and embedded into the HTML file directly to create the animation. I'm not sure that can easily be converted to an actual video file. Without having ffmpeg or avconv installed, that is.

In either case, I agree that we need to tell cate-desktop that the output is an animation.

JanisGailis:
As a side note - I changed the animate_map such that it can animate over an arbitrary dimension. Currently it's a string, but it would be nice if we had a drop-down list of dimensions available to the selected ds - var combination, or at least all dimensions of a ds. So we probably need 'DimNameLike' for that!

forman:
Sounds reasonable. The `DimNameLike` type must be annotated by the variable to which it refers to, so we can fill in the valid value set for the dropdown UI component.

JanisGailis:
Closes #86 #473 #524

JanisGailis:
@forman 

Can you please review this for merging?

Main things to pay attention to:

There's now an HTML return type. This is because this is the only way to produce an animation on all platforms using matplotlib without  packing ffmpeg with cate. With ffmpeg we could output *.mp4, as well as create HTML5 video tags. For *.gif I believe imagemagick is required.

PR #545, opened 2018-03-07
505 mz variable values

forman:


cate/core/workspace.py
```diff
@@ -582,7 +582,7 @@ def run_op(self, op_name: str, op_kwargs: OpKwArgs, monitor=Monitor.NONE):
 
             with monitor.starting("Running operation '%s'" % op_name, 2):
                 self.workflow.invoke(context=self._new_context(), monitor=monitor.child(work=1))
-                op(monitor=monitor.child(work=1), **unpacked_op_kwargs)
+                return op(monitor=monitor.child(work=1), **unpacked_op_kwargs)
```
forman: Note we can only return values that are JSON-serializable. Idea: add kwarg `returns: bool = False` that indicates that we expect a JSON-serializable output rather than side effects (e.g. bringing up a plot, writing a file). Otherwise return `None` (void).

cate/core/wsmanag.py
```diff
@@ -94,7 +94,7 @@ def clean_workspace(self, base_dir: str) -> None:
     @abstractmethod
     def run_op_in_workspace(self, base_dir: str,
                             op_name: str, op_args: OpKwArgs,
-                            monitor: Monitor = Monitor.NONE) -> Workspace:
+                            monitor: Monitor = Monitor.NONE) -> Any:
```
forman: Check: `Union[Any, Void]`  or  `Union[Any, None]`

cate/ops/subset.py
```diff
@@ -89,3 +90,70 @@ def subset_temporal_index(ds: DatasetLike.TYPE,
...
+
+
+@op(tags=['subset', 'utility'], version='1.0')
+@op_input('ds', data_type=DatasetLike)
+@op_input('point', data_type=PointLike)
+@op_input('dim_index', data_type=DictLike, nullable=True)
+@op_return(add_history=False)
+def subset_point(ds: DatasetLike.TYPE,
+                 point: PointLike.TYPE,
+                 dim_index: DictLike.TYPE = None) -> Dict:
```
forman: `dim_index`?  Make name + docs consistent with other operations' index argument.

PR #554, opened 2018-03-09
Kb 546 not possible to download soilmoisture dataset

kbernat:


JanisGailis:
Checking if UC6 runs in this branch. So far so good.

PR #573, opened 2018-03-20
Jg uc11 happy path

JanisGailis:
UC11 notebook and happy paths


JanisGailis:
This pretty much only affects scripts and notebooks folders, so I'm merging it as it has no conflicts with master.

PR #576, opened 2018-03-20
'kb 568 local ds id validation'

kbernat:


cate/ds/local.py
```diff
@@ -644,6 +645,9 @@ def create_data_source(self, data_source_id: str, region: PolygonLike.TYPE = Non
                 meta_info = OrderedDict()
             meta_info['title'] = title
 
+        if not re.match(r'^[\w\-. ]+$', data_source_id):
```
forman: Please comment regex by explaining what it is expected to do, provide examples.

I will use the same in GUI. We may also introduce an operation input annotation that is a regex so we get automatic checks.

forman:
Ok, just comment (see above), then pls. merge.

papesci:
@kbernat please can you close this issue. It seem that the code has been already merged on April

Issue #579, opened 2018-03-21
Xarray >= 0.10.1 time decoding yields unexpected results for SST


JanisGailis:
#548 #443 Are both due to `xarray >= 0.10.1` working differently when decoding time than it used to in `xarray == 0.10.0`. 

See: https://github.com/pydata/xarray/issues/2002

This should be monitored to bump xarray version accordingly when possible. For time being, will pin it to `0.10.0`.

JanisGailis:
The issue has been closed upstream [#2015](https://github.com/pydata/xarray/pull/2015), waiting for release to remove the 10.0 pin.

PR #580, opened 2018-03-21
Pin xarray to 0.10.0 due to #579

JanisGailis:
xarray >= 0.10.1 produces problems with time decoding (at least) for the SST dataset. See #579. Thus, pin the version in environment.yml.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,10 @@
 ## Version 2.0.0.dev8 (in dev)
 
+* SST temporal aggregation error
+  [#548](https://github.com/CCI-Tools/cate/issues/548)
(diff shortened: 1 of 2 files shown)


JanisGailis:
@forman Should't we also add known versioning limitations in ``setup.py``. I didn't do that, as there were no versioning limitations already there, so I figured it must be so for a reason.

forman:
Sorry, forgot to mention: please also update CHANGES.md!

I also need to transfer your changes into `cate-conda`...

JanisGailis:
Feel free to merge this when you see fit! (you have updated cate-conda, or whatever it is that still needs to be done!)

PR #601, opened 2018-04-04
557 pp local ds not visible

papesci:
This branch contain the fix for issue 557 and also a proposal fix for the open_dataset error handling

### DIFF
--- a/cate/util/misc.py
+++ b/cate/util/misc.py
@@ -30,10 +30,42 @@ from contextlib import contextmanager
 from datetime import datetime, date, timedelta
 from io import StringIO
 from typing import Union, Tuple, Sequence, Optional, Iterable
-
+from tornado import httpclient
 import numpy as np
 
 __author__ = "Norman Fomferra (Brockmann Consult GmbH)"
+
+
+def checkUrl(url: str):
+    """
+    This method check wehnever a strng is an url or not
+    it also validate the url, this mean that exists an active
+    remote service and it reply to a get request.
+    In case of failure the error is reported
+    Value error mean in this case that is not an url, may be a file
+    :param url: url string or file path to check
+    :return
+    """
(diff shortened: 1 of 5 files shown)


cate/util/misc.py
```diff
@@ -30,12 +30,44 @@
...
+    remote service and it reply to a get request.
+    In case of failure the error is reported
+    Value error mean in this case that is not an url, may be a file
+    :param url: url string or file path to check
+    :return
+    """
+    http_client = httpclient.HTTPClient()
+    report = {'is_url': True, 'is_valid': True, 'error': None}
+    try:
+        http_client.fetch(url, method='HEAD')
```
forman: This is probably very time consuming call. It is not acceptable as users of the API or the Cate itself may open a large number of datasets.

In fact, I don't think we need a `check_url` method because we need to identify file glob wildcards, see comment above. If at all, what we need a `is_glob` instead.

cate/util/misc.py
```diff
@@ -30,12 +30,44 @@
...
+    This method check wehnever a strng is an url or not
+    it also validate the url, this mean that exists an active
+    remote service and it reply to a get request.
+    In case of failure the error is reported
+    Value error mean in this case that is not an url, may be a file
+    :param url: url string or file path to check
+    :return
+    """
+    http_client = httpclient.HTTPClient()
+    report = {'is_url': True, 'is_valid': True, 'error': None}
```
forman: Be more explicit, you can use 

    report = dict(is_url=True, is_valid=True, error=None)

forman:
@papesci please inform me, when it's time for another review. This PR should be merged ASAP!

papesci:
@forman i made the requested changes in the code adding also a new test. Can you please check it

PR #605, opened 2018-04-06
Open_dataset error hanfdling. 

papesci:
Solve the parameter ambiguity between file name as stings, globs or links
check the existence and availability of the resource local or remote before to create a new xarray


papesci:
@forman you are right it is a repetition of #601. We can remove it

PR #614, opened 2018-04-19
Jg 503 hovmoeller

JanisGailis:
Create Hovmoeller plots.
 
@forman There are new Types now - DimName and DimNamesLike. These are not handled on the GUI, so the operation as it is is not usable in the GUI at the moment.


JanisGailis:
See also: https://github.com/CCI-Tools/cate-desktop/pull/57

PR #625, opened 2018-04-24
update types.py

JanisGailis:
[Mypy](http://mypy-lang.org/) static linker complained in many dozens of places about type incompatibility when ``TYPE`` for ``class Like`` was ``None``. ``Any`` seems to fix this and seems to make more sense than ``None``, unless I misunderstand something. And the errors went away.


cate/core/types.py
```diff
@@ -81,7 +81,7 @@ class Like(Generic[T], metaclass=ABCMeta):
     #: A type that represents the varying source types. This is usually a ``typing.Union`` instance which
     #: combines the varying source types. The ``str`` type shall always be among them so that textual value
     #: representations are supported.
-    TYPE = None
+    TYPE = Any
```
forman: Fine!

PR #626, opened 2018-04-25
393 pp improve error msg

papesci:


forman:
@papesci please write unit-level tests that ensure correct raising of `ValidationError` under exepected conditions.

Then this is ok.

forman:
@papesci Please respond, so we can close this.

forman:
Ok :)

PR #630, opened 2018-04-30
Jg 472 seasonal aggregation

JanisGailis:
``temporal_aggregation`` now is a lot more flexible - works with arbitrary datasets and aggregates to ``month``, ``season`` or ``custom_resolution``, which can be denoted using Pandas [date offset aliases](https://pandas.pydata.org/pandas-docs/stable/timeseries.html#offset-aliases)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -69,3 +69,4 @@ cate/bin
 # Used by Janis (VS Code)
 .mypy_cache/
 .vscode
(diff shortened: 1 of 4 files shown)


JanisGailis:
I'll go ahead and merge this, these are backwards compatible changes to a single operation with associated tests.

Issue #631, opened 2018-05-01
wrong chunk size does not allow to import some dataset 


papesci:
### Expected behavior

Cate should allow to use the following dataset

ID | Dataset Name
--- | ---------------------------------------------------
36 | esacci.LC.13-yrs.L4.WB.asar.Envisat.Map.4-0.r1 
37 | esacci.LC.5-yrs.L4.LCCS.multi-sensor.multi-platform.Map.1-6-1.r1 
44 | esacci.OC.5-days.L3S.OC_PRODUCTS.multi-sensor.multi-platform.MERGED.2-0.r1
46 | esacci.OC.5-days.L3S.RRS.multi-sensor.multi-platform.MERGED.2-0.r1
54 | esacci.OC.8-days.L3S.OC_PRODUCTS.multi-sensor.multi-platform.MERGED.2-0.r1
56 | esacci.OC.8-days.L3S.RRS.multi-sensor.multi-platform.MERGED.2-0.r1
84 | esacci.OC.day.L3S.OC_PRODUCTS.multi-sensor.multi-platform.MERGED.2-0.1997-r1
86 | esacci.OC.day.L3S.OC_PRODUCTS.multi-sensor.multi-platform.MERGED.2-0.1998-r1
88 | esacci.OC.day.L3S.OC_PRODUCTS.multi-sensor.multi-platform.MERGED.2-0.1999-r1
90 | esacci.OC.day.L3S.OC_PRODUCTS.multi-sensor.multi-platform.MERGED.2-0.2000-r1
92 | esacci.OC.day.L3S.OC_PRODUCTS.multi-sensor.multi-platform.MERGED.2-0.2001-r1
94 | esacci.OC.day.L3S.OC_PRODUCTS.multi-sensor.multi-platform.MERGED.2-0.2002-r1
96 | esacci.OC.day.L3S.OC_PRODUCTS.multi-sensor.multi-platform.MERGED.2-0.2003-r1
98 | esacci.OC.day.L3S.OC_PRODUCTS.multi-sensor.multi-platform.MERGED.2-0.2004-r1
100 | esacci.OC.day.L3S.OC_PRODUCTS.multi-sensor.multi-platform.MERGED.2-0.2005-r1
102 | esacci.OC.day.L3S.OC_PRODUCTS.multi-sensor.multi-platform.MERGED.2-0.2006-r1
[...]

Issue #623, opened 2018-04-22
Generate time dim if it doesn't exist


forman:
### Expected behavior

Some NC datasets provide time coverage information using attributes `time_coverage_start`, `time_coverage_end` but don't define a `time` dimension and coordinate variable, e.g. Land Cover CCI.

Cate should generate the missing `time` dimension and coordinate variable if it finds the attributes `time_coverage_start`, `time_coverage_end`.

### Actual behavior

Some operations raise errors if a dataset is lacking a `time` coordinate variable. 

Merging of many such datasets has been addressed in #575 in order to fix #514 but that is not a general solution as it just adds a time dimension, but not a time coordinate variable.

### Specifications 

Cate 1.0 - 2.0.dev9

**EDIT**:

- [x] we must add the time dim when opening datasets from ODP so that timeseries of Land Cover CCI and other ECVs have a time dim

forman:
Reopen as we always want to add a missing time dim when opening datasets.

forman:
Prosper said it's closed now.

PR #632, opened 2018-05-01
Add time dim in open_xarray_dataset() and use file chunking by default

forman:
Hi @JanisGailis, @papesci, please review this asap as it closes

* Generate time dim if it doesn't exist #623
* wrong chunk size does not allow to import some dataset #631


JanisGailis:
It has to be checked if Use Case 6 happy path still works on entry level machines (4GB memory) when using this branch. I'll try to do this tomorrow afternoon.

In addition, it's a good idea to try and download a large subset of SST data and try to do daily -> monthly aggregation on it.

PR #650, opened 2018-05-15
643 nf ensure time is datetime64

forman:


forman:
@JanisGailis you may have a look at this later (I know you are on vacation now)

JanisGailis:
Looks good!

Issue #670, opened 2018-06-08
anomaly_* operations are written only for monthly datasets


JanisGailis:
Both anomaly operations are written and tested for use with monthly datasets. Now that we can aggregate to seasonal datasets these operations should work for at least daily, monthly and seasonal datasets.

- [ ] Write additional tests with daily and seasonal datasets
- [ ] Update the operations to work well with daily and seasonal datasets in addition to monthly datasets

Currently ``anomaly_external`` does more input validation to ensure expected inputs are provided.

PR #672, opened 2018-06-08
666 jg anomaly drops attrs

JanisGailis:
Update ``anomaly_external`` to retain global attributes, update spatial attributes as these may be changed, as well as due more intensive input validation while #670 is open.


JanisGailis:
I'll merge this.

PR #688, opened 2018-06-19
677 removed data_var minimal in open_dataset call

papesci:


papesci:
hi @forman 
    For which issue is this a solution?
the issue number is 677 
    How did you verify the issue (which one?) is solved? Does it work with all dataset types?
i did some test with a separate script to identify the source of the error, Then i have integrated 
the fix into cate running other test using cate cli. In the specific the copy of the dataset reported 
in the issue and then randomly with different dataset, time and variables.
    Provide entry in CHANGES.md!
sorry i have always forgot to update this file.

PR #695, opened 2018-07-04
Pp geo tiff

papesci:
geo tiff reader with optional nromalization


forman:
@papesci what about this PR?

PR #697, opened 2018-07-10
window bug fix in xr opendataset

papesci:
autoclose option when calling xr.open_dataset


forman:
@papesci To which issues does this releate to? As always, please always put an entry in `CHANGES.md` file.

PR #701, opened 2018-07-12
Coerce chunking along time dimension

### LINKED ISSUE #698: Subset spatial with a complex polygon fails when used on locally stored landcover data
From Prosper:

I tried to use a local landcover timeseries data. subset data(using the subset operator) returns data only with meta info( no data actually stored)

-try to subset the landcover data with the Benin shapefile.

-you can replicate this subset with geodata for other datasets and see..

JanisGailis:
The ``read_netcdf`` implementation only provides chunks along spatial
dimensions. This results in all other dimensions including 'time' not
getting chunked. This creates problems later on with retrieving the
computation results for viewing.

The 'time' dimension now gets chunked such that each time
slice is a different chunk, which is a good assumption in
most cases and works well with the given data

Closes #698

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,9 @@
 * Fixed division by zero error in RGB tile generation if data min and max were equal 
 * Allow displaying and working with CCI Sea Level MSLAMPH data.
   Addresses [#531](https://github.com/CCI-Tools/cate/issues/531).
+* Improved chunking when opening local netCDF datasets, improved memory footprint of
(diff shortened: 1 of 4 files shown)


JanisGailis:
@forman Do you think we should provide a ``chunks`` parameter to these ``io`` ``ops``? That way knowledgeable users could experiment with different chunking strategies if something doesn't work and it would be possible to provide more meaningful support in such cases.

PR #715, opened 2018-07-14
684 jg irm coregistration

JanisGailis:
Fixes for coregistration


JanisGailis:
All green, I'll go ahead and merge this, there _should_ be no regressions and in either case it's something contained to a single operation.

PR #724, opened 2018-07-23
windows unverified certificate

papesci:


cate/ds/esa_cci_odp.py
```diff
@@ -170,7 +171,8 @@ def _fetch_solr_json(base_url, query_args, offset=0, limit=3500, timeout=10, mon
             paging_query_args.update(offset=offset, limit=limit, format='application/solr+json')
             url = base_url + '?' + urllib.parse.urlencode(paging_query_args)
             try:
-                with urllib.request.urlopen(url, timeout=timeout) as response:
+                context = ssl._create_unverified_context()
```
forman: Wait. Why does `_create_unverified_context()` start with an underscore, which is therefore a private, non-API method?

PR #731, opened 2018-09-10
Added attributes to _update_resource_json_from_data_frame and _update_resource_json_from_feature_collection to be shown in the WORKSPACE details.

dzelge:


dzelge:
I will need some info what details need to be displayed

forman:
What a about a unit-test?

Issue #734, opened 2018-09-11
cate-desktop unsupported driver GML 


dzelge:
__Expected behaviour__:

Allow or don't allow GML data format when using the read_geo_dataframe operation

__Actual behaviour__:

When chosing GML as input data format I get the follwing error:

```
Traceback (most recent call last):
  File "C:\Users\Helge\AppData\Local\conda\conda\envs\cate-env\lib\site-packages\cate-2.0.0.dev17-py3.6.egg\cate\util\web\jsonrpchandler.py", line 209, in send_service_method_result
    result = future.result()
  File "C:\Users\Helge\AppData\Local\conda\conda\envs\cate-env\lib\concurrent\futures\_base.py", line 425, in result
    return self.__get_result()
  File "C:\Users\Helge\AppData\Local\conda\conda\envs\cate-env\lib\concurrent\futures\_base.py", line 384, in __get_result
    raise self._exception
  File "C:\Users\Helge\AppData\Local\conda\conda\envs\cate-env\lib\concurrent\futures\thread.py", line 56, in run
    result = self.fn(*self.args, **self.kwargs)
  File "C:\Users\Helge\AppData\Local\conda\conda\envs\cate-env\lib\site-packages\cate-2.0.0.dev17-py3.6.egg\cate\util\web\jsonrpchandler.py", line 306, in call_service_method
    result = method(*method_params, monitor=monitor)
  File "C:\Users\Helge\AppData\Local\conda\conda\envs\cate-env\lib\site-packages\cate-2.0.0.dev17-py3.6.egg\cate\webapi\websocket.py", line 291, in set_workspace_resource
    monitor=monitor)
  File "C:\Users\Helge\AppData\Local\conda\conda\envs\cate-env\lib\site-packages\cate-2.0.0.dev17-py3.6.egg\cate\core\wsmanag.py", line 317, in set_workspace_resource
    workspace.execute_workflow(res_name=res_name, monitor=monitor)
  File "C:\Users\Helge\AppData\Local\conda\conda\envs\cate-env\lib\site-packages\cate-2.0.0.dev17-py3.6.egg\cate\core\workspace.py", line 633, in execute_workflow
    self.workflow.invoke_steps(steps, context=self._new_context(), monitor=monitor)
  File "C:\Users\Helge\AppData\Local\conda\conda\envs\cate-env\lib\site-packages\cate-2.0.0.dev17-py3.6.egg\cate\core\workflow.py", line 627, in invoke_steps
[...]

PR #742, opened 2018-09-13
Remove GML option for loading geo data frame

dzelge:
When accepting this PR cate won't have a GML option for loading geo data frames anymore.


dzelge:
This relates to [#734]

Issue #746, opened 2018-09-18
Operation pearson_correlation_scalar() fails


forman:
### What went wrong

```
Job: Executing operation "pearson_correlation_scalar"
Message: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
Code: -32003
Method: set_workspace_resource
Exception: ValueError

Traceback (most recent call last):
  File "C:\Users\Norman\IdeaProjects\ccitools\cate\cate\util\web\jsonrpchandler.py", line 209, in send_service_method_result
    result = future.result()
  File "C:\Users\Norman\Miniconda3\envs\cate-env\lib\concurrent\futures\_base.py", line 425, in result
    return self.__get_result()
  File "C:\Users\Norman\Miniconda3\envs\cate-env\lib\concurrent\futures\_base.py", line 384, in __get_result
    raise self._exception
  File "C:\Users\Norman\Miniconda3\envs\cate-env\lib\concurrent\futures\thread.py", line 56, in run
    result = self.fn(*self.args, **self.kwargs)
  File "C:\Users\Norman\IdeaProjects\ccitools\cate\cate\util\web\jsonrpchandler.py", line 306, in call_service_method
    result = method(*method_params, monitor=monitor)
  File "C:\Users\Norman\IdeaProjects\ccitools\cate\cate\webapi\websocket.py", line 291, in set_workspace_resource
    monitor=monitor)
  File "C:\Users\Norman\IdeaProjects\ccitools\cate\cate\core\wsmanag.py", line 317, in set_workspace_resource
    workspace.execute_workflow(res_name=res_name, monitor=monitor)
  File "C:\Users\Norman\IdeaProjects\ccitools\cate\cate\core\workspace.py", line 652, in execute_workflow
    self.workflow.invoke_steps(steps, context=self._new_context(), monitor=monitor)
  File "C:\Users\Norman\IdeaProjects\ccitools\cate\cate\core\workflow.py", line 632, in invoke_steps
    step.invoke(context=context, monitor=monitor.child(work=1))
  File "C:\Users\Norman\IdeaProjects\ccitools\cate\cate\core\workflow.py", line 318, in invoke
    self._invoke_impl(_new_context(context, step=self), monitor=monitor)
  File "C:\Users\Norman\IdeaProjects\ccitools\cate\cate\core\workflow.py", line 980, in _invoke_impl
    return_value = self._op(monitor=monitor, **input_values)
[...]

PR #753, opened 2018-09-20
Make sure pearson_correlation_scalar input is validated

JanisGailis:
Input validation in pearson_correlation_scalar had faulty logic

### DIFF
--- a/cate/ops/correlation.py
+++ b/cate/ops/correlation.py
@@ -90,7 +90,7 @@ def pearson_correlation_scalar(ds_x: DatasetLike.TYPE,
     array_y = ds_y[var_y]
     array_x = ds_x[var_x]
 
-    if ((len(array_x.dims) != len(array_y.dims)) and
+    if ((len(array_x.dims) != len(array_y.dims)) or
        (len(array_x.dims) != 1)):
         raise ValidationError('To calculate simple correlation, both provided'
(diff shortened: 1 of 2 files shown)


JanisGailis:
Closes #746 

The ``pearson_correlation_scalar`` is meant to work only on 1D variables (much like the underlying scipy function). I had a check for 3D variables, which was faulty. This PR resolves that, but the operation you were trying to carry out in #746 will then fail anyway, just with a nicer message.

JanisGailis:
@forman Merge when ready. If this is not merged in a day, I'll go ahead and do it myself.

forman:
@JanisGailis merging this, but please answer my question once you have time!

JanisGailis:
@forman Using ``.values.flatten()`` on an ND ``xarray.DataArray``, is a bad idea, as it will very likely run into ``MemoryError`` in many instances.

However, after some quick investigations, ``xr.DataArray.stack()`` could be used to achieve what you suggest in a memory safe way. This probably means that the datasets will have to be coregistered beforehand, for the operation to make any sense on ND inputs. This shouldn't take too long to implement. Should I go ahead?

forman:
I actually did not mean the numpy `flatten()`, but just *flatten*, turn ND into 1D. So, yes, fine!

PR #766, opened 2018-09-24
763 nf improve scatter plot

forman:


forman:
True, matplotlib was not made for out of core plotting.

Issue #733, opened 2018-09-10
Coregistration of ocean colour data to SST dataset failing


kjpearson:
### Expected behaviour

Ocean colour should be resampled onto the SST grid after resampling with coregister

### Actual behaviour

"validate" option returns successfully "no problems encountered" but executing the coregister operation fails with: 

Invalid Input

'Oops, Cate couldn't carry out the requested _Executing operation_ "coregister" because:

Could not find a valid intersection to perform coregistration on'

There is then the traceback info but I can't find a way to copy that here. The last call is to coregistration.py, line340, in _find_intersection

### Steps to reproduce the problem

For speed, I have created a subset of the SST CCI L4 v1.1 data for 2004-2005, lat=[-10,10], lon=[-175,-115] to act as the master dataset for coregister. This worked successfully with the cloud cci data as the slave dataset

I have then downloaded a 2004-01-01,2004-05-01 subset of the Ocean Colour CCI dataset "Global attenuation coefficient for downwelling irradiance (Kd490) gridded on a geographic projection V2

Attempting to coregister these two datasets immediately fails because there the ocean colour dataset has a variable "crs" with only a time dimension. I used select_var to extract "MERIS_nobs_sum" ad separately "Kd_940" but trying to coregister either of these caused the error above. I can't see what the problem is as both of these appear to be global datasets.
 
### Specifications 
MacOS, Cate 2.0.0-dev16

kjpearson:
I said above that the cloud cci variables successfully regrid but I'm having second thoughts. Although the coregistration operation completes, it looks like the data for the whole globe in the cloud dataset has been remapped down to the subregion I had in the SST dataset, rather than just those elements of the cloud dataset that were already within the boundary of the subregion. At least, that's how it looks to me from the patterns in the data. The "cee" and "cfc" variables seem the best to look at for this. Could you take a look at this too? If you agree, maybe it should be split to a different issue.

forman:
Hi @JanisGailis could you please get back to @kjpearson

JanisGailis:
@kjpearson I will investigate this. I have seen similar effect before when doing spatial subset, but that turned out to be a visualization issue.

JanisGailis:
@kjpearson The effect with global data being 'compressed' into the subset should be fixed in latest dev release. The problem with finding intersection still persists.

HelenClifton:
[...]

PR #784, opened 2018-09-28
Jg 733 coreg

JanisGailis:
The issue in #733 was that one of the datasets (Ocean Color) had 'inverted' latitude. Previously I managed such cases separately in each operation. It quickly became somewhat troublesome to do that in coregistration. So, I made a decision to take care of inverting the latitude as part of ``normalization``. Based on the following reasons:

1. It is a bit ugly to have to take care of two possible latitude directions in many places
2. It is easy to do once and it works well with ``xarray``, ``dask`` and ``numpy`` providing a view of the original data
3. Although a dataset with decreasing ``lat`` values is a very 'normal' dataset, it's still nice to have harmonized datasets within ``cate``, and other things we normalize for are often also very 'normal' things otherwise.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 
 * Pearson correlation scalar operation now works on N-D data variables [#746](https://github.com/CCI-Tools/cate/issues/746)
 * Make sure integer data variables can be coregistered [#770](https://github.com/CCI-Tools/cate/issues/770)
+* Fix an issue with coregistering datasets with inverted latitudes [#733](https://github.com/CCI-Tools/cate/issues/733)
 
 
 ## Version 2.0.0.dev20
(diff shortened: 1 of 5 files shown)


JanisGailis:
There are some failing tests, but it doesn't seem to be related to what's in this PR!

JanisGailis:
Hi Norman,

Thanks!

There is no ``DataArray.flip()`` that I know of. We do the flipping by 'selecting' the `latitude` dimension in an inverse direction. I'm pretty sure this works with Dask. Mainly because it didn't work with older Dask versions due to some bug. Also, the unit test creates an ``xarray`` dataset with ``dask`` as the underlying array structure. So, yes, it should!

From my tests it also seemed that it just produces a view of the array, so it should work quite well with deferred processing, and should not increase the time it takes to open a dataset much!

forman:
Ok!

PR #794, opened 2018-10-08
343 pp ds name check

papesci:
prevent to create local data source name with unauthorized character

### DIFF
--- a/cate/ds/local.py
+++ b/cate/ds/local.py
@@ -659,8 +659,8 @@ class LocalDataStore(DataStore):
             meta_info['title'] = title
 
         if not re.match(r'^[a-zA-Z0-9_.]*$', data_source_id):
-            raise ValidationError('Unaccepted characters in Data Source name "{}"'.format(data_source_id),
-                                  hint='Use only letters, numbers, dots or underscore in the datasource name')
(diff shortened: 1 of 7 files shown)


cate/ds/local.py
```diff
@@ -658,9 +658,9 @@ def create_data_source(self, data_source_id: str, region: PolygonLike.TYPE = Non
                 meta_info = OrderedDict()
             meta_info['title'] = title
 
-        if not re.match(r'^[\w\-. ]+$', data_source_id):
+        if not re.match(r'^[a-zA-Z0-9_.]*$', data_source_id):
             raise ValidationError('Unaccepted characters in Data Source name "{}"'.format(data_source_id),
```
forman: "Data Source"  --> "data source"

cate/ds/local.py
```diff
@@ -658,9 +658,9 @@ def create_data_source(self, data_source_id: str, region: PolygonLike.TYPE = Non
...
-                                  hint='Do not use space, dot or dash symbol in the datasource name')
+                                  hint='Use only letters, numbers, dots or underscore in the datasource name')
```
forman: "datasource" --> "data source"

Please be consistent.

PR #809, opened 2018-11-08
Jg 805 coreg id grids

JanisGailis:
If datasets provided to `coregistration` have the same geospatial definition, be smarter about the situation. Either return the replica dataset right away, or subset and then return, but don't do the coregistration.


cate/ops/coregistration.py
```diff
@@ -309,6 +314,11 @@ def _resample_dataset(ds_master: xr.Dataset, ds_slave: xr.Dataset, method_us: in
     lat = ds_master['lat'].sel(lat=lat_slice)
     ds_slave = ds_slave.sel(lon=lon_slice, lat=lat_slice)
 
+    if np.array_equal(lon.values, ds_slave['lon'].values) and \
+       np.array_equal(lat.values, ds_slave['lat'].values):
```
forman: Can we put an epsilon, e.g. 1e-8 here?

cate/ops/coregistration.py
```diff
@@ -133,6 +133,11 @@ def coregister(ds_master: xr.Dataset,
                                   ' pixel-registered, can not perform'
                                   ' coregistration'.format(array[0]))
 
+    # Don't do anything if datasets already have the same spatial definition
+    if np.array_equal(ds_master['lat'].values, ds_slave['lat'].values) and \
+       np.array_equal(ds_master['lon'].values, ds_slave['lon'].values):
```
forman: Can we put an epsilon, e.g. 1e-8 here?

cate/ops/coregistration.py
JanisGailis: Yes, let's use:
https://docs.scipy.org/doc/numpy-1.15.1/reference/generated/numpy.allclose.html

JanisGailis:
Turns out this was never merged.

There is some mismatch between master and this branch now. I'll merge this myself, when fixed

forman:
@JanisGailis could you please fix this asap. It's been here since Nov 8 and hence not tested by @HelenClifton.

JanisGailis:
Feel free to merge if the checks pass (not just close it :) ), otherwise, I'll merge it myself when I see that everything has passed.

Issue #823, opened 2018-12-17
Cannot download SST dataset - I/O error


HelenClifton:
### Expected behavior

Tried to download the SST dataset which is used for the use case 6 happy path, using cate GUI.

### Actual behavior

@forman Using cate-2.0.0-dev.23, it has not been possible to download the dataset because there has always been an I/O error before the download has completed. (It has also happened with CLI)
The errors occur at different times after the downlaod has started. An example is below:
![capture](https://user-images.githubusercontent.com/29569434/50081949-a6431c80-01e7-11e9-917d-87441b49b195.PNG)
and here's the traceback: 
[Traceback.txt](https://github.com/CCI-Tools/cate/files/2685562/Traceback.txt)

The last test of the use case 6 was done at the beginning of October using cate-2.0.0-dev.20. At that time , the SST dataset downloaded successfully, although it took about an hour.
Recently though, download speeds have seemed a lot slower (in general, not just this dataset)

### Steps to reproduce the problem

1.  Highlight data source: esacci.SST.day.L4.SSTdepth.multi-sensor.multi-platform.OSTIA.1-1.r1
Click on “Download and/or open remote dataset”

In “Download Data Source” window:
•	Select Time Constraint and enter start time of 2006-01-01 and end time of 2007-12-31

•	Select Region Constraint, enter:
o	Lon. from:  -175
o	Lon. from: -115
o	Lat. from: -10
o	Lat. to: 10

•	Select Variables Constraint, select variables analysed_sst and analysis_error from the list 

•	Select “Download and make local data source“ and enter Unique identifier UC06_SST_2006_2007

•	Click on “Download & Open Local”

2. Wait ! 

### Specifications 

cate-2.0.0-dev.23
Windows 7 Professional

forman:
All these bugs seem to come from a single root cause that applies when we do spatial subsets and consequently use OPeNDAP. Now the list of related issues 

* #822
* #818 
* #816
* #783

forman:
@HelenClifton there is no problem, if you uncheck "Download and make local data source":

![image](https://user-images.githubusercontent.com/206773/50168313-65343080-02eb-11e9-8273-94f5502118ea.png)

Issue #818, opened 2018-11-29
Region of Constraint failure


Evadzi:
@forman @HelenClifton  I closed issue #783 for this new issue

### Expected behavior

I expect cate to download datasets from the ODP with or without specific regions 

### Actual behavior

Cate was able to download data without region of constraints but returns errors for download with region of constraint

NB: CATE RETURNS 2 DIFFERENT ERRORS FOR 2 DIFFERENT DATASETS (OZONE AND LANDCOVER) BUT SAME ISSUE

**e.g 1..ozone... NetCDF: Invalid argument** AND  **e.g 2..Landcover..RuntimeError: NetCDF: Access failure**

EXAMPLE 1
```cate --traceback ds copy esacci.OZONE.mon.L3.NP.multi-sensor.multi-platform.MERGED.fv0002.r1 --name OZ_UC4_1997_2008_GHG --time "1997-01-01,2008-12-31" --region " -160,23,-69,50" 
Sync esacci.OZONE.mon.L3.NP.multi-sensor.multi-platform.MERGED.fv0002.r1: [------------------------------]   0% ESACCI-OZONE-L3-NP-MERGED-KNMI-199701-fv0002.nc                                                    
/home/helge/.cate/data_stores/local/local.OZ_UC4_1997_2008_GHG/ESACCI-OZONE-L3-NP-MERGED-KNMI-199701-fv0002.nc
Sync esacci.OZONE.mon.L3.NP.multi-sensor.multi-platform.MERGED.fv0002.r1: [#-----------------------------]   3% ESACCI-OZONE-L3-NP-MERGED-KNMI-199702-fv0002.nc                                                    
/home/helge/.cate/data_stores/local/local.OZ_UC4_1997_2008_GHG/ESACCI-OZONE-L3-NP-MERGED-KNMI-199702-fv0002.nc
Traceback (most recent call last):
  File "/home/helge/cate/cate/util/cli.py", line 266, in run_main
    args_obj.command_class().execute(args_obj)
  File "/home/helge/cate/cate/util/cli.py", line 163, in execute
    return sub_command_function(command_args)
  File "/home/helge/cate/cate/cli/main.py", line 1311, in _execute_copy
    monitor=cls.new_monitor())
  File "/home/helge/cate/cate/ds/esa_cci_odp.py", line 1011, in make_local
    raise e
  File "/home/helge/cate/cate/ds/esa_cci_odp.py", line 1004, in make_local
[...]

Issue #816, opened 2018-11-29
Fire: Copying remote data failed


Evadzi:
### Expected behavior

Fire data  ingestion from ODP

### Actual behavior

Failed to copy remote data

### Steps to reproduce the problem

1. download fire data using ODP and set region as lon -20, 20   lat  -20, 20

….returns error:-

```Job: Opening data source "esacci.FIRE.day.L4.BA.multi-sensor.multi-platform.MERIS.v4-1.r1"
Message: Data source "esacci.FIRE.day.L4.BA.multi-sensor.multi-platform.MERIS.v4-1.r1": Copying remote data source failed: [Errno -68] NetCDF: I/O failure: b'http://esgf-data1.ceda.ac.uk/thredds/dodsC/esg_esacci/fire/data/burned_area/grid/v4.1/2009/20090207-ESACCI-L4_FIRE-BA-MERIS-fv4.1.nc'
Code: -32003
Method: set_workspace_resource
Exception: cate.core.ds.DataAccessError

Traceback (most recent call last):
  File "C:\Users\prosper\cate-2.0.0.dev22\lib\site-packages\cate\ds\esa_cci_odp.py", line 874, in _make_local
    remote_dataset = adjust_spatial_attrs_impl(subset_spatial_impl(remote_dataset, region),
  File "C:\Users\prosper\cate-2.0.0.dev22\lib\site-packages\cate\core\opimpl.py", line 880, in subset_spatial_impl
    return reset_non_spatial(ds, retset)
  File "C:\Users\prosper\cate-2.0.0.dev22\lib\site-packages\cate\core\opimpl.py", line 782, in reset_non_spatial
    for var_name in ds_source.var():
  File "C:\Users\prosper\cate-2.0.0.dev22\lib\site-packages\xarray\core\common.py", line 54, in wrapped_func
    **kwargs)
  File "C:\Users\prosper\cate-2.0.0.dev22\lib\site-packages\xarray\core\dataset.py", line 2635, in reduce
    **kwargs)
  File "C:\Users\prosper\cate-2.0.0.dev22\lib\site-packages\xarray\core\variable.py", line 1263, in reduce
    data = func(self.data if allow_lazy else self.values,
  File "C:\Users\prosper\cate-2.0.0.dev22\lib\site-packages\xarray\core\variable.py", line 295, in data
    return self.values
  File "C:\Users\prosper\cate-2.0.0.dev22\lib\site-packages\xarray\core\variable.py", line 387, in values
[...]

Issue #783, opened 2018-09-28
Landcover: NetCDF access failure


Evadzi:
### Expected behavior

*I expect cate 2.0.0. dev-20 to successfully ingest landcover data from ODP*

### Actual behavior

*Opening data source failed because of NetCDF access failure*

### error message

 *Job: Opening data source "esacci.LC.5-yrs.L4.LCCS.multi-sensor.multi-platform.Map.1-6-1.r1"
Message: NetCDF: Access failure
Code: -32003
Method: set_workspace_resource
Exception: RuntimeError

Traceback (most recent call last):
 ``` File "C:\Users\prosper\Miniconda3\envs\cate-env\lib\site-packages\cate\util\web\jsonrpchandler.py", line 209, in send_service_method_result
    result = future.result()
  File "C:\Users\prosper\Miniconda3\envs\cate-env\lib\concurrent\futures\_base.py", line 425, in result
    return self.__get_result()
  File "C:\Users\prosper\Miniconda3\envs\cate-env\lib\concurrent\futures\_base.py", line 384, in __get_result
    raise self._exception
  File "C:\Users\prosper\Miniconda3\envs\cate-env\lib\concurrent\futures\thread.py", line 56, in run
    result = self.fn(*self.args, **self.kwargs)
  File "C:\Users\prosper\Miniconda3\envs\cate-env\lib\site-packages\cate\util\web\jsonrpchandler.py", line 306, in call_service_method
    result = method(*method_params, monitor=monitor)
  File "C:\Users\prosper\Miniconda3\envs\cate-env\lib\site-packages\cate\webapi\websocket.py", line 292, in set_workspace_resource
    monitor=monitor)
  File "C:\Users\prosper\Miniconda3\envs\cate-env\lib\site-packages\cate\core\wsmanag.py", line 320, in set_workspace_resource
    workspace.execute_workflow(res_name=res_name, monitor=monitor)
  File "C:\Users\prosper\Miniconda3\envs\cate-env\lib\site-packages\cate\core\workspace.py", line 662, in execute_workflow
    self.workflow.invoke_steps(steps, context=self._new_context(), monitor=monitor)
  File "C:\Users\prosper\Miniconda3\envs\cate-env\lib\site-packages\cate\core\workflow.py", line 627, in invoke_steps
    steps[0].invoke(context=context, monitor=monitor)
[...]

PR #828, opened 2018-12-19
822 forman netcdf4 lock issue

forman:
I'm now using engine='h5netcdf' to store OPeNDAP results because the default engine='netcdf4'
causes crashes in file "netCDF4/_netCDF4.pyx" with currently used netcdf4-1.4.2 conda
package from conda-forge. This occurs whenever remote_dataset.to_netcdf() is called a
second time in the loop that iterates over all OPeNDAP urls found for a query.

Probably related to https://github.com/pydata/xarray/issues/2560.
And probably fixes Cate issues #823, #822, #818, #816, #783.


forman:
@papesci and @dzelge this PR is for your records. I'm merging now.

PR #854, opened 2019-10-22
Updating the cate Python environment

dzelge:
When accepting this PR:

- The environment will be updated and tests that failed with teh update fixed. 
- The aggregation function _lta_general will use ds.mean(). OBS: This may affect downstream analysis procedures as the dimensions have changed.
- A Dockerfile will be added


cate/core/types.py
```diff
@@ -804,7 +804,16 @@ def lazy_data_frame(self):
...
+            df = geopandas.GeoDataFrame.from_features(features, crs=crs)
+            cols = df.columns.tolist()
+            if 'geometry' in cols and cols.index('geometry') != (len(cols) - 1):
+                cols = set(cols) - {'geometry', }
+                cols = list(cols) + ['geometry', ]
+
+                self._lazy_data_frame = df[cols]
+            else:
+                self._lazy_data_frame = df
+
```
forman: Why must geometry be last column?

dzelge: Reverted. there was a test that failrd because geopandas was returning a dataset where the geometry wasn't the last column. But that does not seem to happen anymore.

forman:
All builds failed :(

PR #874, opened 2020-03-05
Suvi cate dirtyfix baseurl k8s

suvarchal:
-  Add ability to change root url path of webapi, so that tornado server responds to requests from that path. This is important for running cate in catehub/jupyterhub context where api service for different users are available on /user/username/. Root url path is deciphered from JUPYTERHUB_SERVICE_PREFIX environment variable.
- Add test to check the above functionality


cate/webapi/start.py
```diff
@@ -98,28 +98,33 @@ def service_factory(application):
...
 def create_application(user_root_path: str = None):
-    application = Application([
-        ('/_static/(.*)', StaticFileHandler, {'path': FigureManagerWebAgg.get_static_file_path()}),
-        ('/mpl.js', MplJavaScriptHandler),
-
-        (url_pattern('/mpl/download/{{base_dir}}/{{figure_id}}/{{format_name}}'), MplDownloadHandler),
-        (url_pattern('/mpl/figures/{{base_dir}}/{{figure_id}}'), MplWebSocketHandler),
+    default_url_root = "/"
+    # replace default url_root with /user/username/ if running in Cate Hub context.
+    url_root = os.environ.get("JUPYTERHUB_SERVICE_PREFIX", default_url_root)
```
forman: Is "JUPYTERHUB_SERVICE_PREFIX" a well-known Juypter environment variable? Or is it your invention?

forman: If it is a well-known Juypter environment variable, then this is fine.
If not, I suggest to call that variable "CATE_WEBAPI_PREFIX".

suvarchal: JUPYTERHUB_SERVICE_PREFIX is Jupyterhub passes as environment variable to each spawned notebook container. we can assign it internally to CATE_WEBAPI_PREFIX but pointless i think.

suvarchal:
> This introduces a WebAPIVersionHandler which is not present elsewhere in the code. Did you forget to commit something?

Thanks, and good catch. it was from merge to master.

dzelge:
You are couple of commits behind master. Expect conflicts or better merge the master into you branch

cate/webapi/start.py
```diff
@@ -98,28 +98,33 @@ def service_factory(application):
...
+    if url_root is not default_url_root:
+        print(f"warning: detected jupyterhub environment variable JUPYTERHUB_SERVICE_PREFIX "
```
TonioF: Why "warning"? Is it a bad thing that this variable is set? I'd prefer "info" or nothing.

suvarchal: Nothing is not good here as environment is detected automatically, info would be good but there is another warning in the module which does similar thing, just retaining it for consistency.

suvarchal:
merged and moving on.

Issue #868, opened 2020-02-04
Utilize new ODP OpenSearch API


TonioF:
The existing odp data store shall be updated so that it uses the newly provided opensearch protocol.

PR #882, opened 2020-03-24
Toniof 868 opensearch

TonioF:
Use new opensearch server as defaut odp data store.
Solves #868 .


cate/ds/__init__.py
```diff
@@ -48,5 +48,8 @@ def cate_init():
     from .esa_cci_odp import set_default_data_store
     set_default_data_store()
 
+    from .esa_cci_odp_os import add_data_store
+    add_data_store()
```
forman: Should be default.

cate/ds/__init__.py
```diff
@@ -48,5 +48,8 @@ def cate_init():
     from .esa_cci_odp import set_default_data_store
     set_default_data_store()
```
forman: Only use for testing, e.g. set a flag `USE_ODP_LEGACY_DATA_STORE = True` or put in config

cate/ds/__init__.py
```diff
@@ -48,5 +48,8 @@ def cate_init():
     from .esa_cci_odp import set_default_data_store
```
forman: raname `esa_cci_odp`   to `esa_cci_odp_legacy`

cate/ds/__init__.py
```diff
@@ -48,5 +48,8 @@ def cate_init():
...
     set_default_data_store()
 
+    from .esa_cci_odp_os import add_data_store
```
forman: rename `esa_cci_odp_os` into `esa_cci_odp`

cate/ds/esa_cci_odp_os.py
```diff
@@ -0,0 +1,1376 @@
...
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+"""
+Description
+===========
+
+This plugin module adds the ESA CCI Open Data Portal's (ODP) OpenSearch service to
```
forman:
```suggestion
This plugin module adds the ESA CCI Open Data Portal's (ODP) service to
```

cate/ds/esa_cci_odp_os.py
```diff
@@ -0,0 +1,1376 @@
...
+the data store registry.
```
forman:
```suggestion
the data store registry. As of April 2020, the ODP service povided a OpenSearch-compatible catalogue,
which is utilised in this implementation.
```

cate/ds/esa_cci_odp_os.py
```diff
@@ -0,0 +1,1376 @@
...
+    def _get_urls_list(files_description_list, protocol) -> Sequence[str]:
+        """
+        Returns urls list extracted from reference esgf specific files description json list
+        :param files_description_list:
+        :param protocol:
+        :return:
+        """
+        return [file_rec[4][protocol].replace('.html', '') for file_rec in files_description_list]
+
+    def _make_local(self,
```
forman: This method is way too large. Split into smaller peaces, so this code becomes human-readable and testable.

forman: Why is this here at all, when there is `make_local(self, ...)` too?

cate/ds/esa_cci_odp_os.py
```diff
@@ -0,0 +1,1376 @@
...
+                                _LOG.info(f"Downloading {actual_url} to {dataset_file}")
+                                urllib.request.urlretrieve(actual_url, filename=dataset_file, reporthook=reporthook)
+                            file_number += 1
+                            local_ds.add_dataset(os.path.join(local_id, filename), (coverage_from, coverage_to))
+
+                            if do_update_of_verified_time_coverage_start_once:
+                                verified_time_coverage_start = coverage_from
+                                do_update_of_verified_time_coverage_start_once = False
+                            verified_time_coverage_end = coverage_to
+        except HTTPError as e:
```
forman: Way to large try/catch scope!

cate/ds/__init__.py
TonioF: Okay. I will consequently add "legacy" from terms that refer to the old data stores and remove "opensearch" to ones that refer to the new one - up to that the identifiers change.

TonioF:
I have addressed the issues from the first iteration.

cate/ds/__init__.py
```diff
@@ -44,9 +44,14 @@ def cate_init():
     # Plugin initializer.
     # Sets the default data store.
 
-    # from .esa_cci_ftp import set_default_data_store
     from .esa_cci_odp import set_default_data_store
     set_default_data_store()
 
+    # from .esa_cci_ftp import set_default_data_store
+    import os
+    if(os.environ.get('USE_ODP_LEGACY_DATA_STORE', False)):
```
forman:
```suggestion
    if os.environ.get('USE_ODP_LEGACY_DATA_STORE', False):
```

PR #890, opened 2020-05-15
use dsr ids to determine data source ids

TonioF:
Use the dsr ids provided by the Open Data Portal to identify data sets. This causes that more datasets are available and others, erroneously included, are removed.

### DIFF
--- a/cate/ds/esa_cci_odp.py
+++ b/cate/ds/esa_cci_odp.py
@@ -761,25 +761,15 @@ class EsaCciOdpDataStore(DataStore):
         self._adjust_json_dict_for_param(json_dict, 'product_string', 'product_strings', values[7])
         self._adjust_json_dict_for_param(json_dict, 'product_version', 'product_versions', values[8])
 
-    def _convert_time_from_drs_id(self, time_value: str) -> str:
-        if time_value == 'mon' or time_value == 'month':
-            return 'month'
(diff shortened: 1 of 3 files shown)


cate/ds/esa_cci_odp.py
```diff
@@ -741,45 +733,54 @@ async def _create_data_source(self, json_dict: dict, datasource_id: str):
...
+            return '5 days'
+        if time_value == '8-days':
+            return '8 days'
+        if time_value == '15-days':
+            return '15 days'
+        if time_value == '13-yrs':
+            return '13 years'
+        if time_value == 'climatology':
+            return 'climatology'
+        raise ValueError('Unknown time frequency format')
```
forman, line 783: Convert the many `if`-s into simple lookup using a dict. 

B.t.w. why is this needed at all?

cate/ds/esa_cci_odp.py
```diff
@@ -955,6 +953,27 @@ def __init__(self,
...
+
```
forman, line 976: Code duplication. See above.

cate/ds/esa_cci_odp.py
TonioF: 1. Better readability. These will be part of the metdata
2. In this notation, they can be used in opensearch queries.

TonioF:
Requested changes are implemented.

Issue #902, opened 2020-07-08
Support data sets with zonal means


TonioF:
The CCI ODP Data Store provides data sets that have no longitude information and instead of a latitude dimension, a dimension named "latitude_centers". These datasets need to be displayed too, preferably as bands spanning a latitude region. To do this, it is necessary to rename "latitude_centers" to "lat" and to insert a new dimension "lon" that covers the globe.

AliceBalfanz:
Implemented by PR #905

PR #905, opened 2020-07-09
Support zonal mean data

AliceBalfanz:
This PR is related to issue #902 .
Cate now creates a longitude dimension for datasets which have zonal means, latitude center and no longitude information.

### DIFF
--- a/cate/core/opimpl.py
+++ b/cate/core/opimpl.py
@@ -80,19 +80,17 @@ def _normalize_zonal_lat_lon(ds: xr.Dataset) -> xr.Dataset:
     :return: a normalized xarray dataset
     """
 
-    if 'latitude_centers' not in ds.coords:
+    if 'latitude_centers' not in ds.coords or 'lon' in ds.coords:
         return ds
-    if 'lon' in ds.coords:
-        return ds
+
     ds_zonal = ds.copy()
     resolution = (ds.latitude_centers.values[1] - ds.latitude_centers.values[0])
     ds_zonal = ds_zonal.assign_coords(lon=[i + (resolution / 2) for i in np.arange(-180.0, 180.0, resolution)])
 
-    for var in ds_zonal.variables:
-        if var not in ds_zonal.coords:
-            if sorted([dim for dim in ds_zonal[var].dims]) == sorted([coord for coord in ds.coords]):
-                ds_zonal[var] = xr.concat([ds_zonal[var] for _ in ds_zonal.lon], 'lon')
-                ds_zonal[var]['lon'] = ds_zonal.lon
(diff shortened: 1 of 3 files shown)


cate/core/opimpl.py
```diff
@@ -71,6 +72,57 @@ def normalize_impl(ds: xr.Dataset) -> xr.Dataset:
...
+
+    if 'latitude_centers' not in ds.coords:
+        return ds
+    if 'lon' in ds.coords:
+        return ds
+    ds_zonal = ds.copy()
+    resolution = (ds.latitude_centers.values[1] - ds.latitude_centers.values[0])
+    ds_zonal = ds_zonal.assign_coords(lon=[i + (resolution / 2) for i in np.arange(-180.0, 180.0, resolution)])
+
+    for var in ds_zonal.variables:
```
TonioF, line 91: So you just want to go through ds_zonal.data_vars? This would be quicker and better readable here.

cate/core/opimpl.py
```diff
@@ -71,6 +72,57 @@ def normalize_impl(ds: xr.Dataset) -> xr.Dataset:
...
+    :return: a normalized xarray dataset
+    """
+
+    if 'latitude_centers' not in ds.coords:
+        return ds
+    if 'lon' in ds.coords:
+        return ds
+    ds_zonal = ds.copy()
+    resolution = (ds.latitude_centers.values[1] - ds.latitude_centers.values[0])
+    ds_zonal = ds_zonal.assign_coords(lon=[i + (resolution / 2) for i in np.arange(-180.0, 180.0, resolution)])
```
TonioF, line 89: I very much dislike this part. We only have the one value, so, it creates a large overhead to copy this ~36 times or so. As currently cate cannot deal with cells with different resolutions, we have no choice, though.

cate/core/opimpl.py
```diff
@@ -71,6 +72,57 @@ def normalize_impl(ds: xr.Dataset) -> xr.Dataset:
...
+    ds_zonal.lon.attrs['long_name'] = 'longitude'
+    ds_zonal.lon.attrs['standard_name'] = 'longitude'
+    ds_zonal.lon.attrs['units'] = 'degrees_east'
+
+    ds_zonal.lat.attrs['bounds'] = 'lat_bnds'
+    ds_zonal.lat.attrs['long_name'] = 'latitude'
+    ds_zonal.lat.attrs['standard_name'] = 'latitude'
+    ds_zonal.lat.attrs['units'] = 'degrees_north'
+
+    ds = ds_zonal.copy()
```
TonioF, line 122: Why copy again? You already did it once above, and even then I didn't think it was necessary.

Issue #904, opened 2020-07-09
Support data sets with time information in file name only


TonioF:
Currently, cate relies on that a time dimension is provided in data sets that shall be concatenated. If time information is only provided in the metadata, a new dimension can be creted from this information. However, Cate can not yet deal with datasets that have their time information in their filenames only.

AliceBalfanz:
Please have another look, the changes cause a problem for of your PR #906 
esacci.OZONE.mon.L3.LP.GOMOS.Envisat.GOMOS_ENVISAT.v0001.r1
esacci.OZONE.mon.L3.LP.GOMOS.Envisat.GOMOS_ENVISAT.v0001.r1
esacci.OZONE.mon.L3.LP.SCIAMACHY.Envisat.SCIAMACHY_ENVISAT.v0001.r1
esacci.OZONE.mon.L3.LP.MIPAS.Envisat.MIPAS_ENVISAT.v0001.r1
esacci.OZONE.mon.L3.LP.SMR.ODIN.SMR_ODIN.v0001.r1
esacci.OZONE.mon.L3.LP.SMR.ODIN.MZM.v0001.r1

AliceBalfanz:
Fixed by TonioF  with Commits on Jul 10, 2020

PR #906, opened 2020-07-09
cate-904 - Support datasets with time information in file name only

TonioF:
Solves #904 .

### DIFF
--- a/cate/core/opimpl.py
+++ b/cate/core/opimpl.py
@@ -33,6 +33,7 @@ from shapely.geometry import box, LineString, Polygon
 from .types import PolygonLike, ValidationError
 from ..util.misc import to_list
 from ..util.monitor import Monitor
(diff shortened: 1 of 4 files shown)


cate/core/opimpl.py
```diff
@@ -327,6 +316,26 @@ def normalize_missing_time(ds: xr.Dataset) -> xr.Dataset:
...
             pass
+    if time_coverage_start or time_coverage_end:
+        return time_coverage_start, time_coverage_end
+    filename = ds.encoding.get('source', '')
+    return get_timestamps_from_string(filename)
+
+
+def normalize_missing_time(ds: xr.Dataset) -> xr.Dataset:
+    """
+    Add a time coordinate variable and their associated bounds coordinate variables
```
AliceBalfanz, line 327: It now covers the case, that the time is only given in the file name, could you add this in the docstring?

PR #908, opened 2020-07-10
Dzelge 14 file download

dzelge:
When accepting this PR the cate web service will have HTTP operations to upload and download datasets


dzelge:
Obsolete

PR #910, opened 2020-07-29
Toniof use xcube store framework

TonioF:
This PR contains a new module esa_cci_odp_xcube.py , in which there is a new DataStore that uses the DataStore from xcube_cci.
The Store is left in its own module on purpose, so it can be compared to the other versions of the ODP stores.
Also note that the new store relies on the latest, non-released versions of xcube and xcube-cci (see also https://github.com/dcs4cop/xcube/pull/329 ).


TonioF:
Outdated. Consider https://github.com/CCI-Tools/cate/pull/990 instead.

PR #913, opened 2020-07-31
Make Polyline/Polygon-Drawing work again

### LINKED ISSUE #912: Current country GeoJSON pathogenic for Cesium globe in Web UI
Error described in https://github.com/CCI-Tools/cate-webui/issues/11.
After many hours of searching, I found the current `cate/ds/data/countries/country.geojson` being the root cause for the misbehaviour of Cesium viewer.

forman:
Closes #912
See also https://github.com/CCI-Tools/cate-webui/issues/11


forman:
> Tested locally, works nicely - one thing: would it be possible to make the area of the countries a bit more transparent. But that is maybe just my personal preference :)

Will do.

PR #916, opened 2020-08-10
New update_file_node() JSON-RPC method

forman:
This PR is required in order run https://github.com/CCI-Tools/cate-webui/issues/64.

It allows updating and scanning of file nodes.

### DIFF
--- a/cate/core/pathmanag.py
+++ b/cate/core/pathmanag.py
@@ -22,7 +22,7 @@ class PathManager:
         normed_path = os.path.normpath(joined_path)
 
         if self._escapes(normed_path):
-            raise ValidationError("Resolves to path outside root path")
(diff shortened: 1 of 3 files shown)


forman:
Requires https://github.com/CCI-Tools/cate/pull/916

forman:
> I am missing Changes and a description, that linux users need to add `--root /home/usr` in order to make the file selector work.

No, this is just a workaround for a current bug. Not intended for users.

> Furthermore, when I navigate to the desired directory, I cannot hit select.

You have to *select* the directory in the dir tree or the table.

> I have to manually enter the path - and '/' does not work - it simply does not show up.

What is the meaning of "/"?

> Having added the correct directory - once added to the the Data Sources, it still gives error message:

I guess, this has nothing to do with the file chooser.

PR #920, opened 2020-08-13
Fix failing make local when no variables or region constraint given

### LINKED ISSUE #919: Failing make local when not passing any variables or region constraints
Saving remote sources to local Store fails when not giving any variables or region constraints:
This results in `HTTP Error 308: Permanent Redirect`
the issue was also described here: 

https://github.com/CCI-Tools/cate-webui/issues/68

AliceBalfanz:
This PR fixes #919 . 
To test:
1. start Cate WebApi server on this branch, 
2. start Cate webui
3. select from remote  `esacci.AEROSOL.5-days.L3C.AEX.GOMOS.Envisat.AERGOM.2-19.r1`
4. go on `Download and/or open remote data source`
image
5. Leave defaults
6. See no error occurring :)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -27,6 +27,7 @@
     - Removed `encoding` keyword argument from `xarray.DataArray` constructor calls.
     - No longer writing to `xarray.DataArray.values` property (e.g. Pearson's
       computation)
+* Downloading of remote data from ODP now handles redirect of url from http to https [#919](https://github.com/CCI-Tools/cate/issues/919)
 
 ## Version 2.0.0
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -27,7 +27,7 @@
     - Removed `encoding` keyword argument from `xarray.DataArray` constructor calls.
     - No longer writing to `xarray.DataArray.values` property (e.g. Pearson's
       computation)
-* Downloading of remote data from ODP now handles redirect of url from http to https [#919](https://github.com/CCI-Tools/cate/issues/919)
+* Downloading of remote data from ODP now handles redirect of URL from HTTP to HTTPS [#919](https://github.com/CCI-Tools/cate/issues/919)
 
 ## Version 2.0.0
 
--- a/cate/ds/esa_cci_odp.py
+++ b/cate/ds/esa_cci_odp.py
@@ -37,6 +37,8 @@ and may be executed using
 Components
 ==========
 """
+import contextlib
+
 import aiofiles
 import aiohttp
 import asyncio
@@ -1253,6 +1255,7 @@ class EsaCciOdpDataSource(DataSource):
                 bytes_to_download = sum([file_rec[3] for file_rec in outdated_file_list])
                 dl_stat = _DownloadStatistics(bytes_to_download)
                 file_number = 1
+                http_redirect = None
                 for filename, coverage_from, coverage_to, file_size, url in outdated_file_list:
                     dataset_file = os.path.join(local_path, filename)
                     child_monitor = monitor.child(work=1.0)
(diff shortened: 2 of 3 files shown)


CHANGES.md
```diff
@@ -27,6 +27,7 @@
     - Removed `encoding` keyword argument from `xarray.DataArray` constructor calls.
     - No longer writing to `xarray.DataArray.values` property (e.g. Pearson's
       computation)
+* Downloading of remote data from ODP now handles redirect of url from http to https [#919](https://github.com/CCI-Tools/cate/issues/919)
```
forman, line 30:
```suggestion
* Downloading of remote data from ODP now handles redirect of URL from HTTP to HTTPS [#919](https://github.com/CCI-Tools/cate/issues/919)
```

cate/ds/esa_cci_odp.py
```diff
@@ -1265,19 +1268,29 @@ def reporthook(block_number, read_size, total_file_size):
                     sub_monitor_msg = "file %d of %d" % (file_number, len(outdated_file_list))
                     with child_monitor.starting(sub_monitor_msg, file_size):
                         actual_url = url[_ODP_PROTOCOL_HTTP]
-                        _LOG.info(f"Downloading {actual_url} to {dataset_file}")
+
+                        if http_redirect:
+                            urlpart = actual_url.split('/', 2)[2]
```
forman, line 1273: Please use `urllib.parse.urlparse()` to get the actual protocol. This will also make sure the URL is valid.

CHANGES.md
forman: Btw, why redirect? Afaik, this a just a protocol adaption.

cate/ds/esa_cci_odp.py
```diff
@@ -1253,6 +1255,7 @@ def _update_ds_using_http(self, local_path, local_ds, selected_file_list, time_r
                 bytes_to_download = sum([file_rec[3] for file_rec in outdated_file_list])
                 dl_stat = _DownloadStatistics(bytes_to_download)
                 file_number = 1
+                http_redirect = None
```
forman, line 1258: Why redirect?

cate/ds/esa_cci_odp.py
```diff
@@ -1265,19 +1268,29 @@ def reporthook(block_number, read_size, total_file_size):
...
+                            urlpart = actual_url.split('/', 2)[2]
+                            actual_url = f'https://{urlpart}'
+                    _LOG.info(f"Downloading {actual_url} to {dataset_file}")
+                    try:
+                        urllib.request.urlretrieve(actual_url, filename=dataset_file, reporthook=reporthook)
+                    except HTTPError as e:
                         try:
+                            actual_url = e.headers['Location']
                             urllib.request.urlretrieve(actual_url, filename=dataset_file, reporthook=reporthook)
+                            http_redirect = True
```
forman, line 1282: How can you know `http_redirect` is now true for other URLs?

cate/ds/esa_cci_odp.py
```diff
@@ -1265,19 +1268,29 @@ def reporthook(block_number, read_size, total_file_size):
                     sub_monitor_msg = "file %d of %d" % (file_number, len(outdated_file_list))
                     with child_monitor.starting(sub_monitor_msg, file_size):
                         actual_url = url[_ODP_PROTOCOL_HTTP]
-                        _LOG.info(f"Downloading {actual_url} to {dataset_file}")
+
+                        if http_redirect:
+                            urlpart = actual_url.split('/', 2)[2]
+                            actual_url = f'https://{urlpart}'
```
forman, line 1274: How can you know, it must be HTTPS? Where does the logic come from?

cate/ds/esa_cci_odp.py
AliceBalfanz: I think, I do not quite understand: 
So the issue is, that the URL used fro request is HTTP, which does exists but is permanently redirected to HTTPS. I could let the code run though 
```
                    except HTTPError as e:
                        try:
                            actual_url = e.headers['Location']
                            urllib.request.urlretrieve(actual_url, filename=dataset_file, reporthook=reporthook)
``` 
but this would mean, that for each file the code first tries http, runs into error and then tries the url form the error header. However I do know, that the requested url belongs to the same dataset collection and therefore I thought I could assume, that if one of the files is https, then all the others would be, too.

Issue #70, opened 2016-10-13
Sketch initial Desktop GUI


forman:
The initial Desktop GUI shall comprise
- a main menu
- multiple tool windows 
  - show available datasets (with filter field)
  - show available operations (with name filter field, allow filtering based on selected dataset/resource)
  - manage current workspace, its workflow and workflow steps (= resources)
  - show details of selected dataset / resource
  - show details of selected operation
- multiple view windows 
  - display selected resource in plot
  - display selected variables as layers of a maps and/or globes
  - allow drawing polygons on map/globe (#72)

Issue #60, opened 2016-09-30
Temporal subset should be able to recognize time in Julian days


JanisGailis:
Currently temporal subset works only with datetime strings. There are other widely used CF compliant time formats, such as time in Julian days, which is used in the clouds CCI. Operations should be able to work with this as well.

mzuehlke:
This applies to the following dataset:
`esacci.CLOUD.mon.L3C.CLD_PRODUCTS.AVHRR.NOAA-15.AVHRR_NOAA.1-0.r1`
The time variable is defined in the following way and not correctly interpreted by xarray:
```
<xarray.DataArray 'time' (time: 12)>
array([ 2454101.5,  2454132.5,  2454160.5,  2454191.5,  2454221.5,
        2454252.5,  2454282.5,  2454313.5,  2454344.5,  2454374.5,
        2454405.5,  2454435.5])
Coordinates:
  * time     (time) float64 2.454e+06 2.454e+06 2.454e+06 2.454e+06 ...
Attributes:
    long_name: time in julian days
    calendar: julian
    units: day
    standard_name: time
```

JanisGailis:
Done.

Issue #56, opened 2016-09-29
Pearson correlation pixel by pixel correlation


JanisGailis:
It should be possible to calculate Pearson correlation on pixel by pixel basis on time/lat/lon datasets

JanisGailis:
Done.

cpaulik:
You might be interested in an recursive implementation of the pearson correlation coefficient. E.g. https://github.com/TUW-GEO/pytesmo/blob/master/pytesmo/metrics.py#L410 to speed things up. You could do it time slice by time slice instead of reading each timeseries using `isel`

PR #922, opened 2020-08-17
Dataset ID appears twice

TonioF:
Solves #56, #60, and #70 .

Actually, the solution to the issues should be that the entry is removed from the ODP. Until then, the entry is not considered when the data sources are constructed. In consequence, all the issues above are solved.

### DIFF
--- a/cate/ds/esa_cci_odp.py
+++ b/cate/ds/esa_cci_odp.py
@@ -747,6 +747,10 @@ class EsaCciOdpDataStore(DataStore):
         for drs_id in drs_ids:
             if drs_id not in self._drs_ids:
                 continue
+            # todo remove as soon as the double occurrence of the drs id is removed from the odp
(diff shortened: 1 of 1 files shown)


cate/ds/esa_cci_odp.py
```diff
@@ -747,6 +747,10 @@ async def _create_data_source(self, json_dict: dict, datasource_id: str):
         for drs_id in drs_ids:
             if drs_id not in self._drs_ids:
                 continue
+            # todo remove as soon as the double occurrence of the drs id is removed from the odp
+            if drs_id == 'esacci.AEROSOL.satellite-orbit-frequency.L2P.AER_PRODUCTS.ATSR-2.ERS-2.SU.4-21.r1' and \
+                    datasource_id == '59f3a38819e140b49ffe46f32176709e':
+                continue
```
forman, line 753: 👍

forman: Not very generic :D

TonioF: Designed to be removed.

PR #923, opened 2020-08-18
renamed legacy ids prefix

TonioF:
Solves https://github.com/CCI-Tools/cate-webui/issues/85 . Later, the Legacy store should be removed.


TonioF:
It worked for me. Can you try deleting your local legacy store? If you still get the error, what is it?

AliceBalfanz:
Ok, I deleted the cached store:
image

TonioF:
I reproduced this. This error is not related to the issue. If I undo my changes and remove the opensearch store, I also get this error. As this has been fixed for the opensearch store, it should be easy to migrate the changes.

Issue #924, opened 2020-08-19
ODP Legacy Store cannot make some data sets local


TonioF:
When it is attempted to call the make_local function on some of the data sets, an error like this is thrown:

Job: Opening data source "cciold.SST.day.L4.SSTdepth.multi-sensor.multi-platform.OSTIA.1-1.r1"
Message: Failed to open data source "cciold.SST.day.L4.SSTdepth.multi-sensor.multi-platform.OSTIA.1-1.r1" for given time range: [Errno -45] NetCDF: Not a valid data type or _FillValue type mismatch: b'https://cci-odp-data.ceda.ac.uk/thredds/dodsC/esg_esacci/sst/data/lt/Analysis/L4/v01.1/2007/01/01/20070101120000-ESACCI-L4_GHRSST-SSTdepth-OSTIA-GLOB_LT-v02.0-fv01.1.nc'
Code: -32003
Method: set_workspace_resource
Exception: cate.core.ds.DataAccessError

This is probably due to recent changes in the ODP.

See here for original post: https://github.com/CCI-Tools/cate/pull/923#issuecomment-675503226

TonioF:
Solved by https://github.com/CCI-Tools/cate/pull/925 .

PR #925, opened 2020-08-19
cate-924: ODP Legacy Store cannot make some data sets local

TonioF:
Solves #924


AliceBalfanz (approved):
Works, nice! thanks for implementing!

Issue #937, opened 2020-08-24
ODP Store keeps producing 500 Errors


TonioF:
When using the ODP Store, it might occur from time to time that users get a message like: https://user-images.githubusercontent.com/42441761/91058800-31192300-e629-11ea-877c-3ebce90315f1.png . In cate-webui, this often happens when data sources are loaded in. These errors often occur although there is content at the requested urls.

TonioF:
Solved by https://github.com/CCI-Tools/cate/pull/938

PR #938, opened 2020-08-24
cate-937: ODP Store keeps producing 500 Errors

TonioF:
Solves #937 . If a 500 error is encountered, it will simply be tried again to access the website, up to 200 times. If the error still exists then, the code will not crash but simply deal with that it gets no result.

### DIFF
--- a/cate/ds/esa_cci_odp.py
+++ b/cate/ds/esa_cci_odp.py
@@ -64,7 +64,7 @@ import xarray as xr
 from cate.conf import get_config_value, get_data_stores_path
 from cate.conf.defaults import NETCDF_COMPRESSION_LEVEL
 from cate.core.ds import DATA_STORE_REGISTRY, NetworkError, DataStore, DataSource, Schema, open_xarray_dataset, \
-    DataStoreNotice
+    DataStoreNotice, DataAccessError, DataAccessWarning
 from cate.core.opimpl import subset_spatial_impl, normalize_impl, adjust_spatial_attrs_impl
 from cate.core.types import PolygonLike, TimeLike, TimeRange, TimeRangeLike, VarNamesLike
 from cate.ds.local import add_to_data_store_registry, LocalDataSource, LocalDataStore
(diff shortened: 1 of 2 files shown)


cate/ds/esa_cci_odp.py
```diff
@@ -278,19 +286,45 @@ def _get_linked_content_from_descxml_elem(descxml: etree.XML, paths: List[str])
...
+            retry_total = retry_min + retry_backoff
+            retry_message = f'Error 429: Too Many Requests. ' \
+                            f'Attempt {i + 1} of {_NUM_RETRIES} to retry after ' \
+                            f'{"%.2f" % retry_min} + {"%.2f" % retry_backoff} = {"%.2f" % retry_total} ms...'
+            warnings.warn(retry_message)
+            time.sleep(retry_total / 1000.0)
+            retry_backoff_max *= _RETRY_BACKOFF_BASE
+        else:
+            break
+    return None
```
forman, line 312: If I see it right, in the end, the actual error information is lost and we'll never know what went wrong. If this is true, please return the HTTP error and status message so users and developers are informed about what went wrong.

cate/ds/esa_cci_odp.py
```diff
@@ -304,21 +338,23 @@ async def _fetch_opensearch_feature_list(base_url, query_args, monitor: Monitor
...
+                if resp:
+                    json_text = await resp.read()
+                    try:
+                        json_dict = json.loads(json_text.decode('utf-8'))
+                        feature_list = json_dict.get("features", [])
+                        full_feature_list.extend(feature_list)
+                        if len(feature_list) < maximum_records:
+                            break
+                    except json.decoder.JSONDecodeError:
+                        print(json_text)
```
forman, line 357: Again, missing error handling. Is this ok, because the JSON's content is not so important?

TonioF:
Edited the error handling.
In most cases when a url could not be retrieved, only a single data source is affected. I altered the code so that a warning is given but the reading of the other data sources is not stopped. If no data could be retrieved, it will not be cached, to prevent that erroneous results will be cached.
The only case when an error will be thrown is when the initial reading of dataset ids has failed.

forman (approved):

Issue #942, opened 2020-09-09
Enable datasets with time format with tailing +00:00


AliceBalfanz:
16 Datasets provided by ODP have their time format stored with a tailing +00:00. 
This currently leads to an error and the datasets cannot be opened: ValueError('unconverted data remains: +00:00')
e.g. esacci.OC.5-days.L3S.RRS.multi-sensor.multi-platform.MERGED.4-2.sinusoidal

See testresults Line 130,  https://github.com/CCI-Tools/cate-e2e/blob/alicja-test-os/validation/python/sorted_test_data_support_os_2020-09-08.csv#L130

AliceBalfanz:
closed by mergin #943

PR #943, opened 2020-09-09
handling time formats with tailing +00:00 #942

AliceBalfanz:
Handle datasets with time format with tailing +00:00 #942


TonioF (changes requested):
Again, no need for a try/except-clause. Just do what is in the except clause always.

cate/ds/esa_cci_odp.py
```diff
@@ -1441,7 +1441,8 @@ async def _init_file_list(self, monitor: Monitor = Monitor.NONE):
         # Compute the data source's temporal coverage
         for file_rec in file_list:
             if file_rec[1]:
-                file_start_date = datetime.strptime(file_rec[1].split('.')[0], _TIMESTAMP_FORMAT)
+                file_start_date = datetime.strptime(file_rec[1].split('/')[0].split('.')[0].split('+')[0],
```
forman, line 1444: The intention of this expression is barely understandable. Leave a comment here to document what's going on ie why this is needed.

cate/ds/esa_cci_odp.py
```diff
@@ -1441,7 +1441,8 @@ async def _init_file_list(self, monitor: Monitor = Monitor.NONE):
...
+                                                    _TIMESTAMP_FORMAT)
                 file_end_date = file_start_date + time_delta
```
forman, line 1446: This may be incorrect, because if the `datetime` is a time range (= has "/" in it) then "<file_start_date>/<file_end_date>", so `file_end_date` is already determined exactly.

forman (changes requested):

cate/ds/esa_cci_odp.py
```diff
@@ -1460,7 +1461,11 @@ async def _init_file_list(self, monitor: Monitor = Monitor.NONE):
         # Compute the data source's temporal coverage
         for file_rec in file_list:
             if file_rec[1]:
-                file_start_date = datetime.strptime(file_rec[1].split('.')[0], _TIMESTAMP_FORMAT)
+                try:
+                    file_start_date = datetime.strptime(file_rec[1].split('.')[0], _TIMESTAMP_FORMAT)
+                except ValueError:
+                    file_start_date = datetime.fromisoformat(file_rec[1].split('.')[0])
```
forman, line 1467: Also `datetime.fromisoformat` may fail with a `ValueError`! Is this handled somewhere else?

AliceBalfanz: Would you expect, e.g.
```
                try:
                    file_start_date = datetime.strptime(file_rec[1].split('.')[0], _TIMESTAMP_FORMAT)
                except ValueError:
                    try:
                         file_start_date = datetime.fromisoformat(file_rec[1].split('.')[0])
                    except ValueError:
                         pass
```
There was no error handling before in this section. I am not sure, what should happen if `ValueError` appears from `datetime.fromisoformat()`

forman (approved):

Issue #944, opened 2020-09-09
Permafrost datasets failing with TypeError("unsupported operand type(s) for -: 'str' and 'str'"))


AliceBalfanz:
The following datasets are failing due to TypeError("unsupported operand type(s) for -: 'str' and 'str'")):
esacci.PERMAFROST.yr.L4.ALT.multi-sensor.multi-platform.MODIS.01-0.r1
esacci.PERMAFROST.yr.L4.GTD.multi-sensor.multi-platform.MODIS.01-0.r1
esacci.PERMAFROST.yr.L4.PFR.multi-sensor.multi-platform.MODIS.01-0.r1

AliceBalfanz:
fixed by #945

PR #945, opened 2020-09-09
Fixed problem with unsupported time format for permafrost datasets #944

AliceBalfanz:
Fixed problem with unsupported time format for permafrost datasets #944


TonioF (changes requested):
There is no need to put get_timestamp_from_string in a try/except-clause. The method takes care of exception handling.

CHANGES.md
```diff
@@ -1,5 +1,6 @@
 ## Version 2.1.1
 
+* Fixed problem with unsupported time format for permafrost datasets #944
```
forman, line 3: Afaik, there is no CCI ECV "permafrost". Please say what datasets are concerned, e.g. provide an ID pattern or so, so people can understand what now work and what was broken..

forman (changes requested):

forman (approved):

Issue #949, opened 2020-09-14
 AttributeError: 'str' object has no attribute 'append' when creating metadata of ODP datasets


AliceBalfanz:
9 AEROSOL Datasets cause great trouble, as they an attribute which is not the expected list! 

test_esa_cci_odp.py::NoDuplicatesTest::test_for_duplicates_in_drs_ids FAILED [100%]
tests/ds/test_esa_cci_odp.py:662 (NoDuplicatesTest.test_for_duplicates_in_drs_ids)
self = <test_esa_cci_odp.NoDuplicatesTest testMethod=test_for_duplicates_in_drs_ids>

    def test_for_duplicates_in_drs_ids(self):
        data_store = EsaCciOdpDataStore()
>       data_sets = data_store.query()

test_esa_cci_odp.py:665: 
                                        
../../cate/ds/esa_cci_odp.py:692: in query
    asyncio.run(self._init_data_sources())
/home/alicja/miniconda3/envs/cate-env/lib/python3.7/asyncio/runners.py:43: in run
    return loop.run_until_complete(main)
/home/alicja/miniconda3/envs/cate-env/lib/python3.7/asyncio/base_events.py:583: in run_until_complete
    return future.result()
../../cate/ds/esa_cci_odp.py:770: in _init_data_sources
    await asyncio.gather(*tasks)
../../cate/ds/esa_cci_odp.py:788: in _create_data_source
    cache_expiration_days=self.index_cache_expiration_days)
../../cate/ds/esa_cci_odp.py:441: in _load_or_fetch_json
    raise e
../../cate/ds/esa_cci_odp.py:426: in _load_or_fetch_json
    json_obj = await fetch_json_function(*(fetch_json_args or []), **(fetch_json_kwargs or {}))
../../cate/ds/esa_cci_odp.py:544: in _fetch_meta_info
    _harmonize_info_field_names(meta_info_dict, 'file_format', 'file_formats')
                                        

catalogue = {'abstract': 'The ESA Climate Change Initiative Aerosol project has produced a number of global aerosol Essential Clim...se data products please see the documentation.', 'bbox_maxx': '180.0', 'bbox_maxy': '90.0', 'bbox_minx': '-180.0', ...}
single_field_name = 'file_format', multiple_fields_name = 'file_formats'
multiple_items_name = None

    def _harmonize_info_field_names(catalogue: dict, single_field_name: str, multiple_fields_name: str,
                                    multiple_items_name: Optional[str] = None):
[...]

PR #950, opened 2020-09-14
fix of weirdo string is not a list issue (duh!) #949

AliceBalfanz:
For some datasets (AEROSOL, 9 datasets) the harmonization of metadata failed due to one attribute not being an expected list but a string. This is now fixed by ensuring, that in case it is not a list, it is converted into one.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 
 * Added package `s3fs` to Python environment as it is required to open Zarr datasets 
   from S3-compatible object store. #940
+* Fixed issue of harmonization of info field names of metadata (#949)
 
 ## Version 2.1.1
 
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -2,6 +2,7 @@
 
 * Added package `s3fs` to Python environment as it is required to open Zarr datasets 
   from S3-compatible object store. #940
+* Fixed issue of harmonization of info field names of metadata (#949)
```
forman, line 5: Please name affected datasets. Use pattern: AEROSOL.*blabla
If this our fault? Or OpenSrearch API? If latter, inform Alison  & Richard .

cate/ds/esa_cci_odp.py
```diff
@@ -374,6 +374,8 @@ def _harmonize_info_field_names(catalogue: dict, single_field_name: str, multipl
         else:
             if catalogue[single_field_name] not in catalogue[multiple_fields_name] \
                     and (multiple_items_name is None or catalogue[single_field_name] != multiple_items_name):
+                if type(catalogue[multiple_fields_name]) != list:
```
forman, line 377:
```suggestion
                if not instanceof(catalogue[multiple_fields_name], list):
```

forman (changes requested):

PR #954, opened 2020-09-15
read_zarr() with credentials

### LINKED ISSUE #953: read_zarr() operation to accept credentials
### Expected behavior

Let `read_zarr()` operation also accept credentials when used with a S3 URL as `path`.

### Actual behavior

 `read_zarr()` forces anonymous object storage access.

### Specifications 

Cate 2.1.2

forman:
Closes #953

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,10 +1,11 @@
 ## Version 2.1.3 (in dev)
 
+* The `read_zarr()` operation now accepts credentials. (#953)
 
 ## Version 2.1.2
 
 * Added package `s3fs` to Python environment as it is required to open Zarr datasets 
-  from S3-compatible object store. #940
+  from S3-compatible object store when using the `read_zarr()` operation. (#940)
 * Fixed issue of harmonization of info field names of metadata (#949)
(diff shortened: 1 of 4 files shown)


AliceBalfanz (changes requested):
Works fine when using s3://some-bucket/some_cube.zarr in aws s3, however when using the url https://s3.eu-central-1.amazonaws.com/some-bucket/some_cube.zarr the cube is not read, shouldn't the second version be supported as well? 

When trying to access a cube stored in the otc cloud obs via 
https://obs.eu-de.otc.t-systems.com/some-bucket/some_cube.zarr this does not work at all. And when I pass s3://some-bucket/some_cube.zarr with file system OBS selected, there is not enough information about which cloud storage should be accessed. 
This should be changed

forman:
You are right, thanks!

forman:
Done.

AliceBalfanz (approved):
Works great! Thanks!

While testing, I noticed, that I cannot cancel the operation `read_zarr` like it is possible e.g. when opening a dataset from CCI. But this might be a separate issue and not part of this PR :)
image

Issue #956, opened 2020-09-21
Include 't' as an indication for time into normalization of cci datasets


AliceBalfanz:
The below datasets have their timestamp indicated by a variable called 't' instead of time - this can be normalized and could include the 4 datasets to be readable by cate. 

esacci.ICESHEETS.yr.Unspecified.GMB.GRACE-instrument.GRACE.UNSPECIFIED.1-2.greenland_gmb_mass_trends
esacci.ICESHEETS.yr.Unspecified.GMB.GRACE-instrument.GRACE.UNSPECIFIED.1-3.greenland_gmb_mass_trends
esacci.ICESHEETS.yr.Unspecified.GMB.GRACE-instrument.GRACE.UNSPECIFIED.1-5.greenland_gmb_mass_trends
esacci.ICESHEETS.yr.Unspecified.GMB.GRACE-instrument.GRACE.UNSPECIFIED.1-4.greenland_gmb_mass_trends

AliceBalfanz:
closed by #959 959

PR #959, opened 2020-09-23
Alicja 956 normalize time called t

AliceBalfanz:
This PR retrieves the time stamps form a dataset in case they are given via the variable 't' instead of 'time'. Before doing so, it checks, whether the values of 't' are in a datetime format or not. This addresses issue #956.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,5 @@
 ## Version 2.1.4 (in development)
-* Included normalization of time in case it is called 't' in a dataset instead of 'time' (#956). 
+* Included normalization of a dataset's time coordinate variable in case it is called 't' instead of 'time' (#956). 
   This is the case for the following datasets: 
     * esacci.ICESHEETS.yr.Unspecified.GMB.GRACE-instrument.GRACE.UNSPECIFIED.1-2.greenland_gmb_mass_trends
     * esacci.ICESHEETS.yr.Unspecified.GMB.GRACE-instrument.GRACE.UNSPECIFIED.1-3.greenland_gmb_mass_trends
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,10 +1,6 @@
 ## Version 2.1.4 (in development)
 * Included normalization of a dataset's time coordinate variable in case it is called 't' instead of 'time' (#956). 
-  This is the case for the following datasets: 
-    * esacci.ICESHEETS.yr.Unspecified.GMB.GRACE-instrument.GRACE.UNSPECIFIED.1-2.greenland_gmb_mass_trends
-    * esacci.ICESHEETS.yr.Unspecified.GMB.GRACE-instrument.GRACE.UNSPECIFIED.1-3.greenland_gmb_mass_trends
-    * esacci.ICESHEETS.yr.Unspecified.GMB.GRACE-instrument.GRACE.UNSPECIFIED.1-5.greenland_gmb_mass_trends
-   This is the case for all datasets with IDs `esacci.ICESHEETS.yr.Unspecified.GMB.GRACE-instrument.GRACE.UNSPECIFIED.1-*.greenland_gmb_mass_trends`.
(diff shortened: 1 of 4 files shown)


CHANGES.md
```diff
@@ -1,5 +1,12 @@
 ## Version 2.1.4 (in development)
-
+* Included normalization of time in case it is called 't' in a dataset instead of 'time' (#956). 
+  This is the case for the following datasets:
```
TonioF, line 3: I don't like listing all the affected dsr id's. You already list them in the linked issue, that should be enough. But let's see what is Norman's opinion on this.

cate/core/opimpl.py
```diff
@@ -383,9 +383,16 @@ def normalize_missing_time(ds: xr.Dataset) -> xr.Dataset:
     if not time_coverage_start and not time_coverage_end:
         # Can't do anything
         return ds
-
+    time = None
     if 'time' in ds:
-        time = ds.time
+        if isinstance(ds.time.values[0], datetime) or isinstance(ds.time.values[0], np.datetime64):
+            time = ds.time
+    elif 't' in ds:
```
TonioF, line 390: Save one line and one indent by concatening lines 390 and 391 with an 'and'

TonioF (approved):
Minor remarks.

cate/core/opimpl.py
AliceBalfanz: No, I had this before with 'and' in one line, then it hit an error whenever time did not exist in ds, but it still tried to check if ds.time.values[0] was an instance of datetime. So I am keeping the current separation.

CHANGES.md
```diff
@@ -1,5 +1,12 @@
 ## Version 2.1.4 (in development)
-
+* Included normalization of time in case it is called 't' in a dataset instead of 'time' (#956).
```
forman, line 2:
```suggestion
* Included normalization of a dataset's time coordinate variable in case it is called 't' instead of 'time' (#956). 
```

CHANGES.md
```diff
@@ -1,5 +1,12 @@
 ## Version 2.1.4 (in development)
-
+* Included normalization of time in case it is called 't' in a dataset instead of 'time' (#956). 
+  This is the case for the following datasets: 
+    * esacci.ICESHEETS.yr.Unspecified.GMB.GRACE-instrument.GRACE.UNSPECIFIED.1-2.greenland_gmb_mass_trends
+    * esacci.ICESHEETS.yr.Unspecified.GMB.GRACE-instrument.GRACE.UNSPECIFIED.1-3.greenland_gmb_mass_trends
+    * esacci.ICESHEETS.yr.Unspecified.GMB.GRACE-instrument.GRACE.UNSPECIFIED.1-5.greenland_gmb_mass_trends
+    * esacci.ICESHEETS.yr.Unspecified.GMB.GRACE-instrument.GRACE.UNSPECIFIED.1-4.greenland_gmb_mass_trends
```
forman, line 7: @TonioF I suggested to @AliceBalfanz listing the dataset IDs here so people can immediately see if their datasets are now supported. However, in this case one could do the following to keep it shorter:

```suggestion
   This is the case for all datasets with IDs `esacci.ICESHEETS.yr.Unspecified.GMB.GRACE-instrument.GRACE.UNSPECIFIED.1-*.greenland_gmb_mass_trends`.
```

cate/core/opimpl.py
```diff
@@ -383,9 +383,16 @@ def normalize_missing_time(ds: xr.Dataset) -> xr.Dataset:
...
+        if isinstance(ds.t.values[0], datetime) or isinstance(ds.t.values[0], np.datetime64):
+            ds = ds.rename_vars({"t": "time"})
+            ds = ds.assign_coords(time=('time', ds.time))
+            time = ds.time
+    if time is not None:
```
forman, line 395: Please document the new behaviour in the docstring.

forman (changes requested):
Only minor changes requested.

Issue #961, opened 2020-09-23
Failing saving to disk with KeyError('geospatial_lat_max'))


AliceBalfanz:
The dataset esacci.SST.day.L4.SSTskin.Unspecified.Unspecified.GMPE.2-0.r1 fails at saving to disk with a 
KeyError('geospatial_lat_max'))

This is due to the retrieval of geospatial information from the remote_dataset's attributes and for that particular dataset this information does not exists. This should not lead to an error as this information may be easily retrieved from the datasets data itself.

TonioF:
Solved by #962

PR #962, opened 2020-09-23
Fix KeyError('geospatial_lat_max'))when saving to disk 

AliceBalfanz:
For saving a dataset locally geospatial information was retrieved from the metadata of the remote dataset. However, in case a coordinate is just one value, it has not been handled well when information about resolution this has now been changed by @TonioF in commit 0137c16 . 

This is adressing issue #961 .


TonioF (changes requested):
I think you fixed it in the wrong place. You achieve the same effect as in your changes if you set allow_point to True in line 1236 . (The problem in your test case is that the subset dataset has a latitude dimension of only 1).
Anyway, the result is not correct in both cases, as it is implied that there is no lat resolution, which is wrong. I think THAT should be fixed.

TonioF (approved):

CHANGES.md
```diff
@@ -1,5 +1,8 @@
 ## Version 2.1.4 (in development)
-
+* When saving a dataset to disk, cate now checks first if the needed keys for geospatial information
```
forman, line 2:
```suggestion
* When caching a dataset to disk, cate now checks first if the needed attributes for geospatial information
```

cate/core/opimpl.py
```diff
@@ -620,6 +620,10 @@ def _get_geo_spatial_cf_attrs_from_var(ds: xr.Dataset, var_name: str, allow_poin
         return None
 
     var = ds[var_name]
+    res_name = 'geospatial_{}_resolution'.format(var_name)
+    min_name = 'geospatial_{}_min'.format(var_name)
+    max_name = 'geospatial_{}_max'.format(var_name)
+    units_name = 'geospatial_{}_units'.format(var_name)
```
forman, line 626: For SST CCI, these attribute may also be named differently:

image

cate/core/opimpl.py
```diff
@@ -650,12 +654,19 @@ def _get_geo_spatial_cf_attrs_from_var(ds: xr.Dataset, var_name: str, allow_poin
...
-                dim_var = var
-                # Actually a point with no extent
-                dim_res = 0.0
-                dim_min = var.values[0]
-                dim_max = var.values[0]
+            elif len(var.values) == 1:
+                if res_name in ds.attrs:
+                    dim_var = var
+                    # Consider extent in metadata if provided
+                    dim_res = float(ds.attrs[res_name].split(' ')[0])
```
forman, line 661: To what does this magic refer to: `ds.attrs[res_name].split(' ')[0]`

forman: If you know it, say so why you split at " ", ignoring everything in the result but the first part.

forman (changes requested):
To me it is not entirely clear why we at all use metadata attributes when we have spatial coordinate variables. Why not always use the correct spatial coordinates from data rather than metadata (which may be wrong, or named differently, etc)?

CHANGES.md
TonioF: Here, we only resort to the information from the metadata when we cannot retrieve the resolution from the data.

cate/core/opimpl.py
TonioF: Right, I'll ,make this clearer. It is not correct as it is anyway.

cate/core/opimpl.py
```diff
@@ -650,12 +654,31 @@ def _get_geo_spatial_cf_attrs_from_var(ds: xr.Dataset, var_name: str, allow_poin
...
-                # Actually a point with no extent
-                dim_res = 0.0
-                dim_min = var.values[0]
-                dim_max = var.values[0]
+            elif len(var.values) == 1:
+                if res_name in ds.attrs:
+                    dim_res = ds.attrs[res_name]
+                    if isinstance(dim_res, str):
+                        # remove any units from string
+                        dim_res = dim_res.replace('degree', '').replace('deg', '').replace('°', '').strip()
```
forman, line 662: What about `'degrees'` or, also commonly used, `'decimal degrees'` and so forth?

As I said, I'd not take this information from metadata attributes. If we have coordinate data, take it from there! There is code in cate or xcube that you can (re)use for this purpose.

forman (changes requested):

cate/core/opimpl.py
TonioF: The reason why we are taking this information from the metadata is because we can not derive it from the coordinates. This is our fallback solution to determine the boundaries in case there is only one entry available for the dimension.

I think it makes sense to state explicitly which names we want to filter out, as sometimes the unit might be 'km' or something else. In that case, I'd rather not use this piece of information.

cate/core/opimpl.py
TonioF: The 'bbox' attribute names are the names as they appear as datasource metadata. Within the xarray datasets, the terms using 'geospatial' are used.

forman (approved):
Thanks!

PR #966, opened 2020-10-12
Only show data sources of the odp data store that can be opened in cate

TonioF:
When we have approved of this, we can do a maintenance release.


AliceBalfanz (approved):
Works nice, thanks!

PR #972, opened 2021-01-14
ensure cate user agent is used in every request

TonioF:
To do this, it was necessary to exchange calls with the urllib library with calls from the requests library. In cases of asyncio, the headers are passed to the session object.

We'd have to figure out a way to pass the user agent to xcube stores, but that is beyond the scope of this pr.


cate/ds/esa_cci_odp.py
```diff
@@ -154,7 +156,7 @@ def find_datetime_format(filename: str) -> Tuple[Optional[str], int, int]:
 
 async def _extract_metadata_from_odd_url(session=None, odd_url: str = None) -> dict:
     if session is None:
-        session = aiohttp.ClientSession()
+        session = aiohttp.ClientSession(headers={'User-Agent': default_user_agent()})
```
forman, line 159: Please extract function / method `aiohttp.ClientSession(headers={'User-Agent': default_user_agent()})` to avoid code duplication.

forman (changes requested):
Please extract function / method `aiohttp.ClientSession(headers={'User-Agent': default_user_agent()})` and use it where appropriate to avoid redundancy and code duplication.

CHANGES.md
```diff
@@ -1,6 +1,7 @@
 ## Version 2.1.5 (in development)
 
 * Now including ODP dataset verification information in data sources for use by Cate App. 
+* Ensure Cate User Agent is set in every http request.
```
forman, line 4:
```suggestion
* Ensure `User-Agent` header is set in every HTTP request made to CCI ODP web APIs. 
```

forman (changes requested):
Ok, thanks! But please write changelog entries _for_ our users.

PR #974, opened 2021-01-14
Dzelge 973 hide secrets

dzelge:
When accepting this PR the cate op `read_zarr` will use the type `Password` for the bucket secret or token


cate/ops/io.py
```diff
@@ -512,8 +512,8 @@ def write_geo_data_frame(gdf: gpd.GeoDataFrame,
 @op_input('drop_variables', data_type=VarNamesLike)
```
forman, line 512: That is the intended way

```suggestion
@op_input('drop_variables', data_type=VarNamesLike)
@op_input('secret', data_type=str, password=True)
@op_input('token', data_type=str, password=True)
```

cate/core/types.py
```diff
@@ -157,6 +157,10 @@ def raise_validation_error(cls, value: Any):
 VarNames = List[str]
 
 
+class Password(str):
+    pass
+
+
```
forman, line 163: Sorry, that is not the intended way.

CHANGES.md
```diff
@@ -1,6 +1,7 @@
 ## Version 2.1.5 (in development)
 
-* Now including ODP dataset verification information in data sources for use by Cate App. 
+* Now including ODP dataset verification information in data sources for use by Cate App.
+* read_zarr is now using type `Password` for the bucket secret or token
```
forman, line 4:
```suggestion
* Operation `read_zarr()` now uses the new `password` attribute its `secret`and `token` parameters.
```

forman (changes requested):
Please use `password` attribute passed to `@op_input` decorator.

TonioF (changes requested):
The entry in changes.md is not understandable. It seems like one word is missing.
Otherwise okay.

forman (approved):

forman (approved):

PR #975, opened 2021-01-14
Apply region constraints before normalizing the time dimension

strawpants:
Dear Cate-devs,

I'm using the cate  on an old machine with little (4gb) ram. When downloading a subset of the land cover data, I ran in to out-of-memory errors because cate tried to somehow load the entire global land_cover array into memory. This seemed odd as I only need a small subset of the data. After some debugging I believe this is triggered by calling xarray' s.expand_dims on the time dimension before any region constraints are applied. After switching the order (applying region constraints before time constraints) it seems to work now. I'm not entirely sure whether this will have consequences, but the locally saved data seemed fine, so I guess it saves bandwidth and time.

Furthermore, I've added a criteria to break out of the loop which saved a dataset to a  netcdf file twice.

Roelof

### DIFF
--- a/cate/core/ds.py
+++ b/cate/core/ds.py
@@ -784,10 +784,12 @@ def open_xarray_dataset(paths,
     if var_names:
         ds = ds.drop_vars([var_name for var_name in ds.data_vars.keys() if var_name not in var_names])
 
-    ds = normalize_impl(ds)
-
     if region:
         ds = subset_spatial_impl(ds, region)
+    
+
+    ds = normalize_impl(ds)
+
 
     return ds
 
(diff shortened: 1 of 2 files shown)


TonioF:
Thank you very much for your pull request. We are very happy that you want to contribute to cate! Your suggested solution to an important issue helps us to develop a solution which does not only work for LC data (as in your PR) but also for other data sets that require that the normalization takes place before a subset is created. We intend to work on a solution which will be in the normalization method itself. We also plan to have other significant changes soon that would override the solution implemented in your PR, and consequently we cannot accept your PR but will address your issue in a future release as described above. I have created an issue for this: https://github.com/CCI-Tools/cate/issues/976

About the other issue you encountered with the data written twice: Thank you for that. I have pushed a fix directly to the master ( https://github.com/CCI-Tools/cate/commit/7463a8f79f886fe75ec0ead95e34805560e24e63 ).

strawpants:
OK great, thanks for addressing this

forman:
We'll revisit this later, when there is a more generic approach in place.

PR #988, opened 2021-04-29
Cate Web API may loose WebSocket connection

### LINKED ISSUE #980: Cate Web API may loose WebSocket connection
### Expected behavior

Cate Web API's WebSocket coinnection should close only if user log outs.

### Actual behavior

Some server configurations (using Nginx) will close WebSocket connections once they are no longer considered alive.
The problem may persist even if the server is configured to allow for very long timeouts, see https://stackoverflow.com/questions/10550558/nginx-tcp-websockets-timeout-keepalive-config.

forman:
Closes #980

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,11 @@
+## Version 3.0.0 (in development)
+
+* Added a Web API function `keep_alive` that is periodically used by Cate App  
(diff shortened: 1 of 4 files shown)


forman:
@dzelge , I cant wait any longer

Issue #939, opened 2020-09-01
Use xcube store framework


TonioF:
The recently designed and developed xcube store framework provides a convenient way to access the data stores that shall be supported by cate and potentially several more. It would be beneficial to use thie framework instead of the cate stores to avoid code redundancy and improve cate's data access capabilities.

TonioF:
ToDo for this

In WebSocketService:
- [ ] Adapt get_data_sources
- [ ] Adapt get_data_stores
- [ ] Implement new method get_data_meta_info (based on get_data_source_temporal_coverage)
In ds.py
- [ ] Adapt ds.open_dataset
In cli.main.py:
- [ ] Adapt _execute_copy
- [ ] Adapt _execute_info
- [ ] Adapt _execute_list

TonioF:
> Adapt ds.open_dataset

Has been implemented but not yet tested

TonioF:
Updated ToDo:

- [ ] Support Vector Data
- [ ] Build in indication which datasets cannot be opened in Cate as Cubes (utilize type specifier)
- [ ] Integrate Progress Monitor from xcube
- [ ] Fix use of asyncio
- [ ] Utilize xcube.datastorepool and/or xcube.datastoreinstance
- [ ] Remove cate ds list --update

TonioF:
Necessary changes to cate app:

[ ] Fetch metadata on demand
[ ] Do not fetch time range anymore

PR #990, opened 2021-05-05
Toniof 939 use xcube

TonioF:
Solves https://github.com/CCI-Tools/cate/issues/939 . In particular, with this PR we replace Cate Data Stores with xcube Data Stores, the DataStoreRegistry with an xcube DataStorePool, and DataSources are removed without an equivalent. DataStores are registered using a configuration file as is used by the xcube DataStorePool.

This PR introduces xcube as a new dependency. Note that to fully exploit the potential of xcube, you will need to install xcube plugins, such as xcube-cci, xcube-cds, and xcube-sh. These plugins are not listed in the environment.yml. Make sure you use at least version 0.8 .


forman (approved):
OMG!

PR #992, opened 2021-05-07
Now recognising cftime.datetime as dtype in time coord

forman:


cate/core/opimpl.py
```diff
@@ -23,17 +23,18 @@
...
 import xarray as xr
 from jdcal import jd2gcal
 from matplotlib import path
 from shapely.geometry import box, LineString, Polygon
 
-from cate.util.time import get_timestamps_from_string, get_timestamp_from_string
 from .types import PolygonLike, ValidationError
 from ..util.misc import to_list
 from ..util.monitor import Monitor
+from ..util.time import get_timestamps_from_string, get_timestamp_from_string
```
TonioF, line 37:
```suggestion
from ..util.time import get_timestamp_from_string
from ..util.time import get_timestamps_from_string
```

TonioF (approved):
I took the liberty to add a test, as I wanted to see the effect of the changes. Please review.

I found that normalize_missing_time erases time coordinates when they are not used as dims and replaces them with values from metadata, which I find wrong. Anyhow, this is beyond the scope of this PR.

forman:
> I took the liberty to add a test, as I wanted to see the effect of the changes. Please review.

Great!

> I found that normalize_missing_time erases time coordinates when they are not used as dims and replaces them with values from metadata, which I find wrong. Anyhow, this is beyond the scope of this PR.

We must fix that. This is the problem with GHG!

Issue #994, opened 2021-05-07
Create sensible names for locally stored data


TonioF:
When data is saved to the local store, it is given an incomprehensible name. Instead, something should be used that allows users to recognize the datasets.

TonioF:
Closed by #996

PR #996, opened 2021-05-07
cate-994: Create sensible names for locally stored data

TonioF:
Solves #994


forman (commented):
Fine! However, I'd prefer if the default for local_name would be the generated from the local_ds_id plus a counter to make sure users don't overwrite their existing cached datasets.

TonioF:
There is a counter.

Issue #929, opened 2020-08-20
Remove requirement for matplotlib <3.3.0


pont-us:
### Expected behavior

With matplotlib 3.3.0 or greater, cate should be able to produce an animated map.

### Actual behavior

With matplotlib 3.3.0 and cartopy 0.17.0, an error results when trying to create an animated map. See Issue #927 and https://github.com/SciTools/cartopy/issues/1615 for details. The obvious fix (specify cartopy >=0.18.0) doesn't work at the time of writing, since it results in an unsolvable conda environment. I haven't been able to track down the exact dependency chain which conflicts with cartopy 0.18.0. As a temporary fix, I restricted matplotlib to <3.3.0, which keeps it compatible with cartopy 0.17.0. In the long term this is not a reasonable solution, since the restricted matplotlib version will become more and more outdated. The package conflict should be tracked down and resolved.

In practice, the issue may resolve itself in time as new package versions with updated dependencies are released.

### Steps to reproduce the problem

1. Attempt to create an animated map via the web UI.
2. Observe the ensuing error. 

### Specifications

Ubuntu 20.04 64-bit, Cate commit 1a1a5d18f21b.

pont-us:
Not clear whether this bug is still occurring.

PR #1000, opened 2021-05-18
Let's see if we can drop matplotlib < 3.3 constraint...

forman:
May close https://github.com/CCI-Tools/cate/issues/929


TonioF (changes requested):
I tested this by following the reproduction procedure from the issue:

- Attempt to create an animated map via the web UI.
- Observe the ensuing error.

This did not work. I got the error message ValueError("cannot use subplot_kws with existing ax") . My first impression is that this is due to the new xarray version.

PR #1004, opened 2021-06-04
Dzelge xxx k8s auto

dzelge:
When accepting this PR cate docker images will:

- be built automatically 
- pushed to the quay docker repository
- and automatically deployed on create release and push to master
- EDIT: unittest will run on GitHub Actions


forman (changes requested):
Move workflow into `bcdev/l8s-configs` so that this repo is not dependent on a particular deployment scenario. The `bcdev/l8s-configs` repo, on the other hand, shall remain self-contained w.r.t. to the deployment scenarios stored in there.

.github/workflows/release_deployments.yml
```diff
@@ -1,27 +1,50 @@
 name: Deploy on Release
```
forman, line 1: It seems that almost all GH actions and workflows use a naming convention, with all lowercase and where hyphens are used instead of underscores. 

If we adhere to that you should rename `release_deployment.yml` into `release-deployment.yml`.

docker/Dockerfile
```diff
@@ -0,0 +1,34 @@
+FROM quay.io/ccitools/cate-base:2.1.0
```
forman, line 1: ~Why putting a single file `dockerfile` into directory `docker`?~

forman: Just found the answer :)

docker/install_xcube_cci.sh
```diff
@@ -0,0 +1,46 @@
...
+  python setup.py install
+
+  cd ..
+  rm v"${XCUBE_CCI_VERSION}".tar.gz
+else
+  mamba update -y -n cate-env -c conda-forge xcube-cci="${XCUBE_CCI_VERSION}"
+fi
+
+# used to ensure all necessary dependencies are included
+mamba install -c conda-forge -y affine click rasterio pydap strict-rfc3339 urllib3 lxml aiohttp nest-asyncio
```
forman, line 46: Add these to Cate's `environment.yml`. Note we need them only if `xcube` and/or `xcube-cci` are installed from source. When from conda, deps come transitive.

forman (commented):
Please see my comment and decide if you'll aply suggested changes.

forman (changes requested):
Please leave a record in `CHANGES.md`.

forman (approved):

PR #1009, opened 2021-06-17
Dzelge xxx docker img deps

dzelge:
When accepting this PR cate's docker image will use a cate environment that is stripped free of any xcube deps allowing 
  to configure the versions of xcube/xcube-cci. These versions are now installed from their respective GitHub release.


bcdev (approved):

forman (commented):
Ok!

PR #1010, opened 2021-06-18
docker: when runnning a cate container it complains that cate is not …

dzelge:
When accepting this PR the cate docker container will not complain anymore about a missing cate package.  When creating the docker image, cate was installed in the python 3.7 lib directory, However, xcube (all of a sudden) change Python to version 3.9 which hid the cate installation.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,12 +18,6 @@ jobs:
           channels: conda-forge
           activate-environment: cate-env
           environment-file: environment.yml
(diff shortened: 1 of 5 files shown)


forman (commented):
No need to change `test.yml` ?

dzelge:
> No need to change `test.yml` ?

Ich habe in test.yml den cate install vorsichtshalber auch nach weiter hinten gestellt.

forman (approved):

PR #1016, opened 2021-09-16
Toniof xxx open local zarrs

TonioF:
Fixes that zarr datasets could not be opened when they were provided by an xcube directory data store.


CHANGES.md
```diff
@@ -1,3 +1,6 @@
+## Version 3.0.1 (in development) 
+* Fixed bug that would cause that no local zarr datasets could be opened.
```
forman, line 2: Please describe the change for the interested user. A "local zarr dataset" ist not what user see. They cannot open cached datasets from the local file store after they have downloaded data from ODP store.

forman (changes requested):

forman (approved):

PR #1020, opened 2021-11-12
Switch to v3.1x

forman:
FYI @AliceBalfanz @dzelge

### DIFF
--- a/cate/ds/__init__.py
+++ b/cate/ds/__init__.py
@@ -38,6 +38,7 @@ information.
 Components
 ==========
 """
+from xcube.util.assertions import assert_given
 
 
 def cate_init():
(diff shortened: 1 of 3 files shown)


cate/ds/__init__.py
```diff
@@ -63,19 +64,24 @@ def cate_init():
             store_configs = yaml.safe_load(fp)
 
     for store_name, store_config in store_configs.items():
-        if store_config.get('store_id', '') == 'file' \
+        store_id = store_config.get('store_id')
+        assert_given(store_id, name='store_id', exception_type=RuntimeError)
```
TonioF, line 68: That's harsh. I think it would be enough to give out a warning and skip the store config.

cate/ds/__init__.py
```diff
@@ -63,19 +64,24 @@ def cate_init():
...
-            base_dir = os.environ.get('CATE_LOCAL_DATA_STORE_PATH',
-                                      os.path.join(get_data_stores_path(),
-                                                   store_name))
-            store_config['store_params']['root'] = base_dir
-        store_params_schema = get_data_store_params_schema(
-            store_config.get('store_id')
-        )
+            root = os.environ.get('CATE_LOCAL_DATA_STORE_PATH',
+                                  os.path.join(get_data_stores_path(),
+                                               store_name))
```
TonioF, line 75: Didn't we agree on that we wanted to get rid of this global variable?

TonioF (changes requested):

cate/ds/__init__.py
forman: This was not goal of this PR. If not set, the default is used, which should now work correctly (with current user "xcube").

cate/ds/__init__.py
forman: I don't agree. The store configuratuion is essential for Cate and a missing `store_id` is not allowed. Hence, Cate must not work with such a wrong configuration.

cate/ds/__init__.py
TonioF: What would be the point of this PR then? To increase the version to 3.1? The change in the changelog has not been considered yet.

forman: Use of xcube 0.9

forman: > What would be the point of this PR then? To increase the version to 3.1?

Yes. See title of PR.

cate/ds/__init__.py
forman: Ideally, the JSON/YAML validation should take place after reading.

Issue #1022, opened 2021-12-01
Adding Local Data Sources fails when called from Cate App


TonioF:
When you select "Add Data Source" for a mutable data store in Cate App, you will get the error message "No paths found", even if there is data at the specififed path, satisfying the specified file pattern.

A quick investigation showed that this happens because in the Cate Websocket the path is not correctly resolved.

TonioF:
Solved by https://github.com/CCI-Tools/cate/pull/1023

PR #1023, opened 2021-12-01
Cate-1022: Adding Local Data Sources fails when called from Cate App

TonioF:
Solves #1022


forman (approved):
Thanks!

forman:
...but pls fix unit-.tests first! :)

forman (approved):

PR #1025, opened 2021-12-08
Fixed logging and improved auto-stop behaviour

forman:
* Cate logger is now configured independently from the Tornado logger (now logging works!)
* Changed behavior of option `--auto-stop-after` of `cate-webapi-start`  CLI tool: 
  - WebSocket messages will now reset the auto-stop timer too. 
  - Using `time.time()` instead of `time.perf_counter()` to compute idle time


TonioF (approved):
Works!

PR #1027, opened 2021-12-09
Notify time left until auto-close

### LINKED ISSUE #1026: Notify service clients about upcoming auto-stop
### Expected behavior

Users of the Cate App backend service should be informed about upcoming auto-shutdown due to inactivity, given that auto-shutdown has been enabled for the service (e.g. `cate-webapi-start --auto-stop-after 3600 ...`).

### Actual behavior

The service stops without notification just leaving a log message.

1. Run `cate-webapi-start --auto-stop-after 100 ...` on localhost.
2. Start Cate App for local service. 
3. In Cate App do nothing else, just wait for "Connection closed" message (see https://github.com/CCI-Tools/cate-app/issues/150).

### Implementation

Use WebSocket to send JSON RPC message to clients containing time left until auto-close service and time of service inactivity.

### Specifications 

Cate 3.1.1, Cate App 3.1.0

forman:
The "keep_alive" WebSocket JSON RPC method of the `cate-webapi-start` CLI tool will now return an object with auto-close info, if `cate-webapi-start` is started with the `--auto-stop-after` option. Otherwise it returns `null`. The object comprises the fields:

- `available_time` - configured auto-stop time in seconds.
- `inactivity_time` - duration in seconds of service inactivity.
- `remaining_time` - duration in seconds remaining until service auto-stop.

closes #1026

**EDIT**

This is a prerequisite for https://github.com/CCI-Tools/cate-app/pull/165


AliceBalfanz (approved):

cate/webapi/start.py
```diff
@@ -97,7 +97,8 @@ def get(self):
 
 
 def service_factory(application):
-    return WebSocketService(application.workspace_manager)
+    return WebSocketService(application.workspace_manager,
+                            application)
```
TonioF, line 101:
```suggestion
    return WebSocketService(
        application.workspace_manager,
        application if application.auto_stop_after else None
    )
```

TonioF (changes requested):
I keep getting errors when I do not pass autostop as argument to cate.  See my comments for a possible fix.

cate/webapi/start.py
forman: Thanks, great bug!

TonioF (approved):

PR #1028, opened 2021-12-15
Removed outdated occurrence of directory

TonioF:
Extremely minor commit where the oudated store_id 'directory' is replaced with the new store_id 'file'

### DIFF
--- a/cate/webapi/websocket.py
+++ b/cate/webapi/websocket.py
@@ -172,7 +172,7 @@ class WebSocketService:
             config = DATA_STORE_POOL.get_store_config(instance_id)
             data_stores.append(dict(id=instance_id,
                                     title=config.title,
-                                    isLocal=config.store_id == 'directory',
+                                    isLocal=config.store_id == 'file',
                                     description=config.description,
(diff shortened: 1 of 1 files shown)


cate/webapi/websocket.py
```diff
@@ -172,7 +172,7 @@ def get_data_stores(self) -> List[Dict[str, Any]]:
             config = DATA_STORE_POOL.get_store_config(instance_id)
             data_stores.append(dict(id=instance_id,
                                     title=config.title,
-                                    isLocal=config.store_id == 'directory',
+                                    isLocal=config.store_id == 'file',
```
forman, line 175: Why do you call this an "extremely minor commit". Cate App treats such stores writable, hence will alow users to add datasets --> crash.

forman: It also seems there is no unit test that would have indicated the bug.

TonioF: Because we have the opposite siutation: As no store is considered local, CateApp will allow users not to add datasets to any store.
Also, in the current implementation of CateApp, the only instance where this is considered is when the offline mode is checked.

Issue #1030, opened 2022-01-20
Trying to overwrite previously datasets causes in a DataStoreError


TonioF:
### Expected behavior

When you create a local copy of a cate dataset, it should be saved under the specified name. If there already is a dataset with the same name in a user's local store, that dataset should be overwritten.

### Actual behavior

The dataset is not overwritten. Instead, you get an error like: 

xcube.core.store.error.DataStoreError: Failed to write dataset '/home/xcube/.cate/data_stores/local/name.zarr': path '' contains a group

### Steps to reproduce the problem

1. Select a dataset from the CCI Store
2. Open it and select Cache Data Source option, set an identifier
3. Repeat steps 1 and 2, use the same identifier

TonioF:
Closed by #1031

PR #1031, opened 2022-01-20
Allow re-using identifiers to cache data

TonioF:
Solves #1030


TonioF:
If you want all checks to be succesful, #1032 must be accepted and merged into this PR.

forman (approved):

PR #1032, opened 2022-01-20
Avoid using python 3 10

TonioF:
Avoid using Python 3.10 as long as there are problems with transitive dependencies.
Also, there is one adapted import.


TonioF:
For this, we need to wait for the next xcube release first.

forman (approved):

Issue #1034, opened 2022-03-09
API docs not populated


forman:
Although https://cate.readthedocs.io/en/latest/api_reference.html includes complete sections for the API Reference, none of these sections have been (automatically) populated with any content. Similarly, section 2.3 (API Usage) of the Developer Guide is also empty.

PR #1035, opened 2022-03-10
Repopulated API Docs

TonioF:
Solves #1034 . 
What I did:
- added missing mock imports
- fixed some layout issues that would cause errors during the doc creation
- Updated entries (removed references to functions/packages/modules that no longer exist, addedreferences to new ones)

You can see the resulting new API Docs here: https://cate.readthedocs.io/en/toniof-1034-docs/


forman (approved):
Thanks!

PR #1036, opened 2022-03-11
Added missing doc chapters

TonioF:
Solves #1034 .


docs/source/developer_guide/dg_api_usage.rst
```diff
@@ -1,2 +1,137 @@
...
+This will give you a dictionary where the keys are the operation names and the values are information about the
+operations, including their keyword arguments.
+
+To apply an operation, you can do this:
+
+.. code-block:: python
+
+    from cate.core.op import OP_REGISTRY
+    op = OP_REGISTRY.get_op(op_name)
+    op(**op_kwargs)
```
forman, line 110: This is the generic use case. It is still good to have it here, but most users will import the desired operation from the `cate.ops` package directly. Please say so in the beginning of this section.

forman (changes requested):
Very good, thanks a lot! 

Please refer to my comments for the requested changes.

forman (approved):
Great!

PR #1040, opened 2022-12-05
Extended operation `no_op` by parameter `memory_alloc` for container

forman:
Addresses #980


TonioF (changes requested):
The memory is not allocated in each step, only once. Either adjust the parameter description or move the memory.append()-line into the for loop.

TonioF (approved):

Issue #1041, opened 2023-03-01
Writing feature collection fails


TonioF:
The method `write_feature_collection` fails with `TypeError: Object of type Feature is not JSON serializable` in some cases.
You can best reproduce this by running the test `test_countries_with_simp` in `WriteFeatureCollectionTest`.

PR #1042, opened 2023-03-01
cate-1041: Writing feature collection fails

TonioF:
Solves #1041 . You might need to update you environment to get the problem described in the issue first.


TonioF:
Update: To get the tests run, I had to update the appveyor script and change in a setting in an otherwise unrelated test.

forman (approved):
Thanks!

PR #1045, opened 2023-03-25
cate 4.0 - workspace id instead of base dir

forman:
Now using workspace identifiers instead of base directories in resource URLs of the WebAPI. This way we no longer need to URL-encode workspace directories in WebAPI URLs, which did not work with [jupyter-server-proxy](https://jupyter-server-proxy.readthedocs.io/).


TejasMorbagal (approved):
Looks good to me

TonioF (approved):

PR #1048, opened 2023-03-31
Adjust workspaces directories

forman:
In this PR:

* Workspaces directories are now under "${root}/.cate" or "~/.cate".
* Scratch directory is now cleared before being used.
* Added cate-app 4.0.0-dev.5 (from https://github.com/CCI-Tools/cate-app/pull/198)


TonioF (approved):
Looks cleaner than before and I like that the scratch is cleaned (this might lead to some undesired behaviour / data loss, but users can be encouraged to always use workspaces so their work is saved).
I don't yet see though how this ensures that the scratch and workspace directories are created in the home directory.

forman:
I'll merge although unit tests fail. But they fail due to reasons outside of this PR.

